PR #8, opened 2016-04-06
Add tests

aspidites:
# Description

This PR (once complete) adds tests for functionality, which should help guide creation of new backends. I determined which tests to write by following the tutorials
# Progress

I'll update this list soon to include all tests I plan to write to make it easier to track progress of this PR.
- [x] table creation
- [ ] row insertion

### DIFF
--- a/beam.cabal
+++ b/beam.cabal
@@ -46,7 +46,7 @@ test-suite spec
   hs-source-dirs:     test
   main-is:            Spec.hs
   build-depends:      base >=4.7 && <4.9, text, beam, hspec, QuickCheck
-  extensions: StandaloneDeriving, TypeSynonymInstances,FlexibleInstances, TypeFamilies, DeriveGeneric, OverloadedStrings
+  extensions:         StandaloneDeriving, TypeSynonymInstances,FlexibleInstances, TypeFamilies, DeriveGeneric, OverloadedStrings
 
 source-repository head
   type: git
--- a/beam.cabal
+++ b/beam.cabal
@@ -2,7 +2,7 @@
 -- see http://haskell.org/cabal/users-guide/
 
 name:                beam
-version:             0.4.0.0
+version:             0.4.1.0
 synopsis:            A type-safe SQL mapper for Haskell that doesn't use Template Haskell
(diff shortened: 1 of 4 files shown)


beam.cabal
```diff
@@ -2,7 +2,7 @@
 -- see http://haskell.org/cabal/users-guide/
 
 name:                beam
-version:             0.4.0.0
+version:             0.4.1.0
```
aspidites: I tried to follow PVP here. We didn't change the type signature of anything, and only added new functionality, so the minor version was bumped.

src/Database/Beam/Backend.hs
```diff
@@ -116,11 +116,17 @@ openDatabase' isDebug db strat dbSettings =
...
 
+
 dumpSchema :: Database db => DatabaseSettings db -> IO ()
-dumpSchema (db :: DatabaseSettings db) =
+dumpSchema (db :: DatabaseSettings db) = do
+  putStrLn "Dumping database schema ..."
+  schema <- showSchema db
+  mapM_ putStrLn $ lines schema
+
+showSchema :: Database db => DatabaseSettings db -> IO String
```
aspidites: As dumpSchema returned IO (), I couldn't capture the result, which I needed for tests.

aspidites:
@tathougies can you offer any guidance into if there is a simple way to capture the sql generated by a given command. I see that debug options allow those to go to console, but I'd prefer a pure function that just spit out potential SQL, if possible.

tathougies:
@aspidites Sorry for the long delay! Had to take a break for personal reasons :)

In terms of getting the command for a particular expression.. There's currently no way to do that, but it would be straightforward to add. If you look in Database/Beam/Query.hs, you'll see all the functions we can use to query/update the database. Each of these (query, insertInto, updateWhere, etc) has an associated function (queryToSQL', insertToSQL, updateToSQL) respectively which returns a SQLSelect, SQLInsert, or SQLUpdate respectively. You can use these objects to construct a SQLCommand, which can be pretty printed using the Database.Beam.Query.ppSQL function.

Let me know which functions you think you'd need, and I can take a stab at them.

aspidites:
Thanks. I'll give it ago myself and ping you if I have trouble. May be a
minute. Going through my own personal stuff now so remaining motivated has
been... difficult as of late.

PR #10, opened 2016-06-10
Refactoring and Improvements to the Example

hesiod:
This PR contains a variety of refactoring and again fixes a lot of warnings. Just like with the last PR, this is quite monolithic, but if you want me to split off some parts of, just say so.

Interesting changes (citing mostly from 51b75a0):
* In some places, statements of the form `\x -> case x of { a -> b; ... }` have been replaced by `\case { a -> b; ... }`, which is syntactic sugar enabled by the LambdaCase extension. LambdaCase has been added to the default-extensions field in the cabal file.
* In some places, type variables in annotations in the function body obviously intended to refer to type variables in the top-level annotation, in the spirit of ScopedTypeVariables, but without the explicit forall required by ScopedTypeVariables to bring top-level type variables into scope in the function body. In other words, there were annotations like `x :: a` in the function body which are meaningless without a `forall a. a -> ...` in the function annotation.

Other than that, I restructured and refactored the example quite a bit to make it easier to grasp.


hesiod:
I added some more refactoring commits just now.

hesiod:
...and finally, some more changelog entries

tathougies:
Thanks for these.... Everything looks good to me!

tathougies:
Do you have any other fixes, or can I merge these in?

hesiod:
I've got more commits in the works, but I'd rather have them in a separate PR because they are more experimental (i.e. possibly controversial) than this, so go ahead! And thanks for merging!

tathougies:
Looks good to me!

PR #12, opened 2017-03-05
Add missing version bounds, make bounds DRY

mightybyte:
Cabal only needs the version bounds specified once, which makes things less error prone in the future.


tathougies:
Thanks!

PR #18, opened 2017-04-14
Fix "Ambiguous Occurrence 'C'"

ramirez7:
```
Database/Beam/Schema/Tables.hs:28:17: error:
    Ambiguous occurrence ‘C’
    It could refer to either ‘GHC.Generics.C’,
                             imported from ‘GHC.Generics’ at Database/Beam/Schema/Tables.hs:61:1-40
                          or ‘Database.Beam.Schema.Tables.C’,
                             defined at Database/Beam/Schema/Tables.hs:264:1
```

looks like it was introduced by 0c1ee6992e0d852c303313a7e529a24df9742d62


tathougies:
Oh good catch!

PR #32, opened 2017-07-15
Static check for column types

gip:
Prevent invalid column field types. Only basic types have an instance for now, other types can be added in the backend.

```haskell
data SomeTableT f
  = SomeTable
  { field1  :: Columnar f Text       -- Ok
  , field2  :: Columnar f Int        -- Ok
  , field3  :: Columnar f (Int, Int) -- Wrong, should be rejeted
  } deriving Generic
```


tathougies:
Field validity is backend specific. Beam itself is happy to shuffle data around regardless of what the underlying type is. Moreover, tuples are used in some backends to handle row values (part of SQL92), so `Columnar f (Int, Int)` is a potentially valid type in every backend.

PR #37, opened 2017-08-03
Relax QIf* expressions to allow contexts other than QValueContext.

luigy:
This allows queries like the following to have aggregates in the condition:
```sql
select (case when max(foo) > 5 then 'a' else 'b' end) from generate_series(1,2) foo;
```
``` 
case 
------
 b
(1 row)
```


tathougies:
Perfect, thanks!

PR #45, opened 2017-08-21
Add `Day` support to beam-migrate and beam-sqlite.

judah:


beam-sqlite/Database/Beam/Sqlite/Syntax.hs
```diff
@@ -644,3 +644,7 @@ instance HasSqlValueSyntax SqliteValueSyntax ByteString where
 instance HasSqlValueSyntax SqliteValueSyntax LocalTime where
   sqlValueSyntax tm = SqliteValueSyntax (emitValue (SQLText (fromString tmStr)))
     where tmStr = formatTime defaultTimeLocale (iso8601DateFormat (Just "%H:%M:%S%Q")) tm
+
+instance HasSqlValueSyntax SqliteValueSyntax Day where
```
tathougies: You'll likely need `FromBackendRow` for both `Postgres` and `Sqlite`. Do they both have instances (they may, I don't know)?

judah: Yep, looks like they both have it:
https://github.com/tathougies/beam/search?utf8=%E2%9C%93&q=frombackendrow+day&type=

I confirmed that my code using `Day` compiled with this change.

PR #49, opened 2017-09-20
Allow aeson 0.11

### LINKED ISSUE #48: beam-postgres: Widen aeson bound to include 0.11
GHCJS is held at `aeson 0.11` and in order to maintain safe communication between client and server those of us using it also want to pin `aeson` to 0.11 on the server. However, beam-postgres doesn't allow for it. Looking at the history I don't see any particular reason not to try it.

3noch:
Fixes #48


tathougies:
Would you mind adding yourself to CONTRIBUTORS?

3noch:
Haha, I don't normally imagine that bumping version bounds qualifies, but as you wish!

tathougies:
@3noch I often don't have the mind for this kind of maintenance, so I appreciate all the help I can get!

3noch:
Thanks!

PR #51, opened 2017-09-22
Fix local inference issue with GHC 8.2

ElvishJerricco:
This appears to be a regression in GHC, but I can't for the life of me figure out how to reproduce it.


tathougies:
Huh... interesting!

PR #54, opened 2017-09-26
Add null ordering

luigy:
https://ronsavage.github.io/SQL/sql-2003-2.bnf.html#null%20ordering

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL/SQL2003.hs
+++ b/beam-core/Database/Beam/Backend/SQL/SQL2003.hs
@@ -4,6 +4,7 @@ module Database.Beam.Backend.SQL.SQL2003
     ( module Database.Beam.Backend.SQL.SQL99
 
     , IsSql2003FromSyntax(..)
+    , IsSql2003OrderingElementaryOLAPOperationsSyntax(..)
     , IsSql2003ExpressionSyntax(..)
     , IsSql2003ExpressionElementaryOLAPOperationsSyntax(..)
     , IsSql2003ExpressionAdvancedOLAPOperationsSyntax(..)
@@ -33,6 +34,11 @@ class IsSql92FromSyntax from =>
                     -> Maybe Integer
                     -> Maybe Text
                     -> from
+
+-- | Optional SQL2003 "Elementary OLAP operations" T611 support
+class IsSql92OrderingSyntax ord =>
+    IsSql2003OrderingElementaryOLAPOperationsSyntax ord where
+    nullsFirstOrdering, nullsLastOrdering :: ord -> ord
 
 class ( IsSql99ExpressionSyntax expr
       , IsSql2003WindowFrameSyntax (Sql2003ExpressionWindowFrameSyntax expr) ) =>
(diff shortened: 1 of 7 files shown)


beam-core/Database/Beam/Backend/SQL/SQL2003.hs
```diff
@@ -34,6 +35,10 @@ class IsSql92FromSyntax from =>
                     -> Maybe Text
                     -> from
 
+class IsSql92OrderingSyntax ord =>
+    IsSql2003OrderingSyntax ord where
+    nullsFirst, nullsLast :: ord -> ord
```
tathougies: To be in line with the naming convention in `IsSql92OrderingSyntax` can we rename these `nullsFirstOrdering` and `nullsLastOrdering`?

beam-core/Database/Beam/Backend/SQL/SQL2003.hs
```diff
@@ -34,6 +35,10 @@ class IsSql92FromSyntax from =>
                     -> Maybe Text
                     -> from
 
+class IsSql92OrderingSyntax ord =>
+    IsSql2003OrderingSyntax ord where
```
tathougies: I believe this ordering syntax is part of the extension T611: Elementary OLAP Operations. Would you mind renaming the class `IsSql2003OrderingElementaryOLAPOperationsSyntax` and adding a comment indicating the extension supported: 

```haskell
-- | Optional SQL2003 "Elementary OLAP operations" T611 support
```

beam-postgres/Database/Beam/Postgres/Syntax.hs
```diff
@@ -373,6 +373,10 @@ instance IsSql92OrderingSyntax PgOrderingSyntax where
   ascOrdering e = PgOrderingSyntax (fromPgExpression e <> emit " ASC")
   descOrdering e = PgOrderingSyntax (fromPgExpression e <> emit " DESC")
 
+instance IsSql2003OrderingSyntax PgOrderingSyntax where
```
tathougies: One thing I worry about here is what happens if you do `nullsFirst (nullsLast asc)`. I think that would generate incorrect sql. I think we could fix this by modifying `PgOrderingSyntax` to be a data type instead of a newtype.

So something like

```haskell
data PgOrderingSyntax
  = PgOrderingSyntax
  { pgOrderingSyntax :: PgSyntax
  , pgOrderingNullOrdering :: Maybe PgNullOrdering
  }
```

And then define

```haskell
fromPgOrdering :: PgOrderingSyntax -> PgSyntax
fromPgOrdering (PgOrderingSyntax s Nothing) = s
fromPgOrdering (PgOrderingSyntax s (Just PgNullsFirst)) = s <> emit " NULLS FIRST"
```

Or something like that? I just want to make sure something sensible happens.

tathougies:
Would you add a note to `compatibility.md` in the docs? Otherwise, looks good to me!

docs/user-guide/queries/advanced-features.md
```diff
@@ -37,6 +37,20 @@ This extension also provides various window functions for SQL. The only one beam
...
 appreciated!
 
+### Null Ordering
+
+This optional SQL2003 feature allows nulls to appear before or after non-null
+values in the sort ordering.
+
+!beam-query
+```haskell
+!chinook sqlite3
```
tathougies: Not supported on sqlite :) Remove this line, if possible.

PR #56, opened 2017-10-02
Support variable projection of fields

luigy:


tathougies:
Looks good to me. Have you tested it?

PR #57, opened 2017-10-03
updating docs for installing until released on hackage

jacobono:
Just adding some docs to use in a project since beam isn't realeased to stackage yet.

just use the stack file to pull directly from master to whichever commit you
would like.

After that, add beam-core, beam-migrate, and the backend of your choice
[postgres|sqlite] (to your cabal file)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -43,8 +43,9 @@ packages:
     - <backend>
 ```
 
-NB: the commit will need to be changed to whatever the latest commit of master
-is, or whichever commit you want to build from even.
+!!! note "Note"
+    the commit will need to be changed to whatever the latest commit of master is,
+    or whichever commit you want to build from even.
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -27,21 +27,34 @@ Beam features
...
+    commit: a3b5e0763843fed48c7eef53fa7d08cfe710342d
+  extra-dep: true
+  subdirs:
+    - beam-core
+    - beam-migrate
+    - <backend>
 ```
 
-or
+NB: the commit will need to be changed to whatever the latest commit of master
```
tathougies: Use the `!!! note` syntax here.

jacobono:
added the note syntax -- I forgot about that syntax when I first started.  Fixed now.

tathougies:
Thank you very much!

PR #58, opened 2017-10-03
WIP separating tutorials for each backend

jacobono:
sqlite is moved to `tutorials/sqlite`
postgres is moved to `tutorials/postgres`

tutorial 3 isn't done yet since I am still getting errors on the custom marshalling bit

I could probably do another pass through to make sure all of my notes and stuff are included in the first two tutorials.

I just yanked and pasted from the project you were looking at with some of the problems.


kmicklas:
Closing this due to inactivity and conflicts.

Feel free to fix the conflicts and reopen. I do think it could be a good idea to tailor the tutorials to backend specifics, but I'd prefer if there was a way to do that without copy and pasting huge chunks of text.

PR #60, opened 2017-10-05
Add postgres array_length and array_agg quantifier variant

luigy:


beam-postgres/Database/Beam/Postgres/PgSpecific.hs
```diff
@@ -149,6 +149,23 @@ arrayLowerUnsafe_ (QExpr v) (QExpr dim) =
   QExpr . PgExpressionSyntax $
   emit "array_lower(" <> fromPgExpression v <> emit ", " <> fromPgExpression dim <> emit ")"
 
+arrayLength_
```
tathougies: Could you prefix with `pg` like the other functions?

luigy: Oh yeah I meant to ask you that as well since array_upper/lower also didn't prefix with `pg`?

PR #61, opened 2017-10-06
Widen dlist lower bound

3noch:
Things appear to work with wider `dlist` bounds. I needed this because `reflex-platform`'s `ghcjs` package set is currently stuck on dlist 7.2.1. If you'd rather not merge this in order to minimize your compatibility "footprint", I can work around it easily on my end. But I figured I'd offer it as a change in case you want it.


tathougies:
@3noch did you mean `0.7.2.1`? The constraint you added doesn't make sense?

3noch:
Oops. Yes it should have been `0.7.1.2`. Fixed.

PR #67, opened 2017-10-11
Bumped base version bounds for beam-migrate

LightAndLight:
`beam-migrate` builds with base-4.10.0.0, so I increased the version bounds to reflect this


tathougies:
Great... thank you!

PR #72, opened 2017-10-16
Add UPDATE..RETURNING support for postgres

luigy:


beam-postgres/Database/Beam/Postgres/Syntax.hs
```diff
@@ -1028,6 +1031,29 @@ insertReturning (DatabaseEntity (DatabaseTable tblNm tblSettings))
          in emit " RETURNING "<>
             pgSepBy (emit ", ") (map fromPgExpression (project (mkProjection tblQ))))
 
+newtype PgUpdateReturning a = PgUpdateReturning PgSyntax
+
+updateReturning :: Projectible PgExpressionSyntax a
```
tathougies: Perhaps get rid of the `Maybe` for `returning`?

luigy: oh totally that is the only way this is used

tathougies:
Thanks again @luigy !

PR #74, opened 2017-10-16
postgres json[b] needs to be single quoted for postgres to not complain

jacobono:
I noticed this trying to insert a jsonb into a table

with debug mode on, I can see the sql generated as

```sql
INSERT INTO "some_table"
("some_jsonb") 
VALUES 
([{"task":"do the other thing"}]::jsonb), ([{"task":"do this other thing"}]::jsonb)
```

```text
*** Exception: SqlError {sqlState = "42601", sqlExecStatus = FatalError, sqlErrorMsg = "syntax error at or near \"[\"", sqlErrorDetail = "", sqlErrorHint = ""}
```

and perusing the [postgres json docs](https://www.postgresql.org/docs/9.6/static/datatype-json.html) I could see that beam is missing the single quotes

```sql
SELECT '{"bar": "baz", "balance": 7.77, "active":false}'::json;
```

I'm using postgres `9.6.5`, and this fixes the errors for me.  Wondering if you can check on your machine?

Also I know there aren't any tests for a change like this... would you want to start small with tests for these changes and provide some guidance on how to do so?

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/PgSpecific.hs
+++ b/beam-postgres/Database/Beam/Postgres/PgSpecific.hs
@@ -221,7 +221,7 @@ instance (Typeable a, FromJSON a) => FromBackendRow Postgres (PgJSON a)
 instance ToJSON a => HasSqlValueSyntax PgValueSyntax (PgJSON a) where
   sqlValueSyntax (PgJSON a) =
     PgValueSyntax $
-    escapeString (BL.toStrict (encode a)) <> emit "::json"
+    emit "'" <> escapeString (BL.toStrict (encode a)) <> emit "'::json"
 
 newtype PgJSONB a = PgJSONB a
   deriving ( Show, Eq, Ord, Hashable, Monoid )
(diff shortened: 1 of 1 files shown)


tathougies:
Yes, beam is almost definitely missing single quotes here. I feel something must be off, because I'm using it in production for a side project

beam-postgres/Database/Beam/Postgres/PgSpecific.hs
```diff
@@ -221,7 +221,7 @@ instance (Typeable a, FromJSON a) => FromBackendRow Postgres (PgJSON a)
 instance ToJSON a => HasSqlValueSyntax PgValueSyntax (PgJSON a) where
   sqlValueSyntax (PgJSON a) =
     PgValueSyntax $
-    escapeString (BL.toStrict (encode a)) <> emit "::json"
+    emit "'" <> escapeString (BL.toStrict (encode a)) <> emit "'" <> emit "::json"
```
tathougies: Collapse these two `emits` into one `emit "'::json"`. Otherwise, looks good.

tathougies:
Ah I see, it's fixed in the `nextgen-beam-migrate` branch but wasn't backported.

jacobono:
glad it was there in another branch!  also collapsed those two emits like you asked.  I wasn't sure if it was more clear to have the single quote emissions to show the escaping or smash the last quote in like you mentioned.

would it be worthwhile to have some tests that check the PgValueSyntax on things like this?

tathougies:
Tests would be nice, but we don't have any tests for the postgres backend right now. I'll get to it later, but PRs always welcome

PR #99, opened 2017-12-18
Add support to rename all fields

### LINKED ISSUE #89: Rule based table field modifications
I use a different scheme for naming records than what Beam supports. Because of this I have to override almost all of my table fields. This gets old fast. However, if I could provide a function to specify how to do the renaming, I could remove all of the boilerplate. How hard would it be to allow the renaming rule to be overridden at a DB or table level?

tathougies:
Fixes #89

### DIFF
--- a/beam-core/Database/Beam/Schema/Tables.hs
+++ b/beam-core/Database/Beam/Schema/Tables.hs
@@ -24,7 +24,7 @@ module Database.Beam.Schema.Tables
     , withTableModification, modifyTable, fieldNamed
     , defaultDbSettings
 
-    , RenamableWithRule(..), Renamablefield(..)
+    , RenamableWithRule(..), RenamableField(..)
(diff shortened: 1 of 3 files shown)


tathougies:
@3noch You wanted this :) There's an example in the tests: 

https://github.com/tathougies/beam/blob/master/beam-core/test/Database/Beam/Test/Schema.hs#L218

3noch:
@tathougies Wow fantastic!!

moll:
If I'm not entirely mistaken, I think the following today does run on column names ahead of Beam's mangling:

```haskell
cacheDb :: Beam.DatabaseSettings Beam.Sqlite CacheDb
cacheDb =
  Beam.defaultDbSettings
  `Beam.withDbModification`
  Beam.renamingFields (yourRenamerHere . NonEmpty.head)
```

PR #113, opened 2018-01-17
Merge nextgen-beam-migrate

tathougies:


mckeankylej:
Woooo! Super happy to see this branch merged!

tathougies:
@mckeankylej Great! Just FYI, this does not mean the `beam-migrate` tool should be considered production ready or complete, but it's at the point where the branches are getting out of sync and merging them is hassle.

PR #120, opened 2018-01-26
Correctly generate projection when joining against an ORDER BY over a filtered SELECT

### LINKED ISSUE #70: filtering an orderBy causes invalid SQL instead of type error
I had a query like this:

```haskell
runSelectReturningList $ select $ limit_ 5 $ filter_ cond $ orderBy_ order $ all table
```

This produced incorrect SQL.

Switching the `filter_` and `orderBy_` clauses works as expected.

tathougies:
Fixes #70


3noch:
Great! Thanks for looking into this!

tathougies:
@3noch There are a few other corner cases I've uncovered, and there'll be fixes for them soon too. I'll tag you in the PR... thanks for reporting this!

PR #136, opened 2018-02-11
Add lower_ and upper_

mulderr:
Seems like low hanging fruit. Copied `octetLength_` with minor changes. Hope it's ok. Basic tests pass with both postgres and sqlite.


tathougies:
Great! Thank you very much!

PR #139, opened 2018-02-12
Allow up to 12-tuples in results

mulderr:
There seems to be a limit of 8 on the size of tuples that can be returned as results. I ran into this by accident with a query that returns a 9-tuple. This adds instances needed to handle up to 12.

Don't need to go there just yet but all things considered could this be a valid exception from the non-TH rule?


tathougies:
As it is, you should be able to nest tuples arbitrarily deep. If you need 12 fields, then just nest a 6 tuple inside a 2 tuple, or a 3 tuple inside a 4 tuple. Did that not work for you?

mulderr:
Ah, that just didn't occur to me. I need to see how that plays out. I'll close this and get back to you if I have more trouble. Thanks!

PR #155, opened 2018-02-26
Postgres SELECT ... FOR UPDATE Support

SamProtas:
Referencing: https://github.com/tathougies/beam/issues/152

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Syntax.hs
+++ b/beam-postgres/Database/Beam/Postgres/Syntax.hs
@@ -1435,13 +1435,13 @@ pgDropExtensionSyntax extName =
 --   https://www.postgresql.org/docs/10/static/sql-select.html#SQL-FOR-UPDATE-SHARE
 --   "The locking clauses cannot be used in contexts where returned rows cannot be clearly
 --   identified with individual table rows; for example they cannot be used with aggregation."
-withLocking_ :: (Database db)
-             => PgSelectLockingStrength
-             -> [DatabaseEntity Postgres db (TableEntity table)]
-             -> Maybe PgSelectLockingOptions
-             -> Q PgSelectSyntax db s a
-             -> Q PgSelectSyntax db s a
-withLocking_ lockStrength tbls mLockOptions q = undefined
+--withLocking_ :: (Database db)
+--             => PgSelectLockingStrength
+--             -> [DatabaseEntity Postgres db (TableEntity table)]
+--             -> Maybe PgSelectLockingOptions
+--             -> Q PgSelectSyntax db s a
+--             -> Q PgSelectSyntax db s a
+--withLocking_ lockStrength tbls mLockOptions q = undefined
 
 now_ :: QExpr PgExpressionSyntax s LocalTime
 now_ = QExpr (\_ -> PgExpressionSyntax (emit "NOW()"))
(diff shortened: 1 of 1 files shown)


SamProtas:
I _think_ I've got the various locking clause forms covered. Although I was unsure of how to specify the table names for the `OF <table1>[,<table2>,...]` exactly.

SamProtas:
It's likely I'm missing some things, but at this point I'm thinking the following:

- We can't use `buildSql92Query` anymore for postgres' `HasQBuilder` instance
    - I'm guessing this is "correct" in the sense that this functionality isn't in the 92 standard
    - Maybe some (most) of this convenience function can be factored out to be reusable for a `buildPostgreSqlQuery` function 
    - If my next thought below is true, maybe a `buildQueryExtensions` function can be passed in to share most of this logic still
- `QF` may need to be extended in some way. I'm not sure the best way to do this
    - Maybe we could add another type parameter (defaulted to `()`) and constructor for backends to included a nested free-monad defining their additons (I'm not 100% sure this is even possible).

Or maybe there's something much simpler I haven't figured out.

beam-postgres/Database/Beam/Postgres/Syntax.hs
```diff
@@ -1371,6 +1421,23 @@ pgDropExtensionSyntax extName =
 --                                        Nothing
 --      buildJoinFrom tbl newSource Nothing
 
+
+-- TODO
+-- Don't directly expose sum-type constructors?
```
tathougies: They should be exposed from the `Syntax` module, but they should not be exposed from the main `Database.Beam.Postgres` module, if that makes sense.

beam-postgres/Database/Beam/Postgres/Syntax.hs
```diff
@@ -1371,6 +1421,23 @@ pgDropExtensionSyntax extName =
...
+-- Constrain 'table' to exist within the query (using s maybe?)
+-- Constrain Q to not have aggregations (at least the locked tables I think, using s?)
+--
+--   https://www.postgresql.org/docs/10/static/sql-select.html#SQL-FOR-UPDATE-SHARE
+--   "The locking clauses cannot be used in contexts where returned rows cannot be clearly
+--   identified with individual table rows; for example they cannot be used with aggregation."
```
tathougies: Oh this is an interesting restriction. I'll have to think what the best way to deal with this one is. We could probably do it with the `s` parameter, but may need to make some changes in `beam-core`.

beam-postgres/Database/Beam/Postgres/Syntax.hs
```diff
@@ -1371,6 +1421,23 @@ pgDropExtensionSyntax extName =
 --                                        Nothing
 --      buildJoinFrom tbl newSource Nothing
 
+
+-- TODO
+-- Don't directly expose sum-type constructors?
+-- Constrain 'table' to exist within the query (using s maybe?)
```
tathougies: I think we can probably do some trickery with existential types on `db` to make sure this works. Let me think about it.

tathougies:
@SamProtas Overall, I'm happy accepting this pull request as is and then work on these issues one by one. If you could just modify it so the constructors are exposed, that'd be great.

SamProtas:
Let me know if you had something else in mind.

tathougies:
Perfect! Thank you very much. Please add yourself to the CONTRIBUTORS file, if you've not already done so!

PR #171, opened 2018-03-17
Add `createSchema` shortcut

### LINKED ISSUE #154: Creating DB schema on the first run?
So, I know there is migration library, but I'm wondering if it's possible to use it to create all the required tables in an empty database. Right now, from what I understand, I have to create the initial schema by using SQL / DDL and beam can only do CRUDE operations on it.

Am I missing something?

tathougies:
Fixes #154


tathougies:
This is a work-in-progress.

PR #173, opened 2018-03-19
Implemented locking clause for postgres

### LINKED ISSUE #152: SELECT .... FOR UPDATE Support Planned?
I couldn't find anything in the docs/code to indicate this was currently possible or considered yet. Is this feature planned?

Postgres in particular has a lot of useful variations of `lock strength` and `NOWAIT / SKIP LOCKED`.

tathougies:
Fixes #152

@SamProtas Let me know what you think of the syntax. I've given some examples in the docs.

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Full.hs
+++ b/beam-postgres/Database/Beam/Postgres/Full.hs
@@ -7,25 +7,36 @@
 -- "Database.Beam.Query" and provide a strict superset of functionality. They
 -- map 1-to-1 with the underlying Postgres support.
 module Database.Beam.Postgres.Full
-  ( lockingAllTablesFor_, lockingFor_
-  , PgWithLocking
-  , PgLockedTables, locked_, lockAll_, withLocks_
-  , PgSelectLockingStrength(..), PgSelectLockingOptions(..)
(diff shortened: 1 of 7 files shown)


beam-postgres/Database/Beam/Postgres/Full.hs
```diff
@@ -1,22 +1,100 @@
...
+instance Monoid (PgLockedTables s) where
+  mempty = PgLockedTables []
+  mappend (PgLockedTables a) (PgLockedTables b) = PgLockedTables (a <> b)
+
+data PgWithLocking s a = PgWithLocking (PgLockedTables s) a
+instance ProjectibleWithPredicate c syntax a => ProjectibleWithPredicate c syntax (PgWithLocking s a) where
+  project' p mutateM (PgWithLocking tbls a) =
+    PgWithLocking tbls <$> project' p mutateM a
+
+-- | Use with `withLocking_` to lock all tables mentioned in the query
```
SamProtas: I think this should be `lockingFor_` right?

tathougies: Yes.. great catch!

PR #181, opened 2018-03-24
Export autoMigrate from Database.Beam.Migrate.Simple

joshvera:
So users can access it from other modules. Based off of https://github.com/tathougies/beam/issues/154#issuecomment-373897308.


tathougies:
Thank you!

PR #189, opened 2018-03-28
Be more strict about when to accept notNull

### LINKED ISSUE #177: Use of `maybeType` and `notNull` together in migration
The beam snapshot I am using is `6a80a598073` (The last one pushed on 20 March).

I have noticed that in my migration function, it is possible to have a field like
```haskell
(field "my_row" (maybeType myType) notNull)
```
which says a row is both nullable and not-nullable, which doesn't make much sense, and causes a run-time exception when "my_row" is `Nothing`. Of course, writing something like this is a mistake from the part of the developer :stuck_out_tongue: .

I was wondering if it can be guaranteed at compile time that no-one does something like this? Would be more of an enhancement, I imagine.

tathougies:
Fixes #177

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,5 @@ install:
 
 script:
 # Build the package, its tests
-- stack --no-terminal --flag beam-core:werror --flag beam-sqlite:werror --flag beam-postgres:werror --flag beam-migrate:werror test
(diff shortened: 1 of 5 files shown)


tathougies:
Note to self: travis CI builds failed because the build file configuration was wrong.

PR #190, opened 2018-03-28
Bump stack resolver

SamProtas:
I was interested in upgrading a project to LTS 11 and was curious what it would take bring Beam up.

This is the minimal PR to get it to compile.

Lower bounds:
I didn't touch them, but some of these changes (at least `haskell-src-exts`) aren't backwards compatible. If you were to accept this PR at some point, how would you like to handle this? Bumping Beam's version and bumping the lower bounds? CPP?

DerivingStrategies:
I haven't fully looked into how beam-migrate uses `haskell-src-exts` but I tried to keep the previous behavior (no deriving strategies clauses).

Conduit:
Conduit dropped finalizers. I think removing the only use of them shouldn't be a big deal, but I'm not an expert with Conduit.


tathougies:
I will take a look. thanks!

PR #193, opened 2018-04-03
Remove functional dependencies `be -> m`, `handle -> m` on MonadBeam

kmicklas:
While it is convenient for inference, these dependencies make it impossible to naturally use a monad transformer on MonadBeam.

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL.hs
+++ b/beam-core/Database/Beam/Backend/SQL.hs
@@ -33,7 +33,7 @@ import Control.Monad.IO.Class
 --   are supported in individual backends. See the documentation of those
 --   backends for more details.
 class (BeamBackend be, Monad m, MonadIO m, Sql92SanityCheck syntax) =>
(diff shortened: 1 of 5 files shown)


tathougies:
If you can make the build pass, I will likely accept this PR.

kmicklas:
Thanks! It seems like there are still CI failures due to warnings that are unrelated to my changes.

3noch:
@kmicklas I've noticed this as well. Thanks for looking into it.

ocharles:
How much inference do you lose if you take these out?

kmicklas:
@ocharles It shouldn't matter unless you are trying to infer the monad of a `MonadBeam` constraint from the backend or handle types. (I think this is pretty obscure and something you shouldn't do anyway.)

tathougies:
@kmicklas you're right, the errors are unrelated. Thanks for #194, I'll merge these two this morning

tathougies:
Thank you!

PR #195, opened 2018-04-04
Discussion point & proof of concept: Have you considered using Backpack?

ocharles:
Hi Travis!

First off, thanks for writing such a fantastic library. You may be aware that I've also been writing [my own database library - Rel8](https://github.com/ocharles/rel8). I spent the long bank holiday here in the UK doing a fairly drastic rewrite of it - removing arrow syntax, mainly. Rel8 encourages the use of Beam-like HKTs, but unlike Beam it doesn't mandate them, and has some slightly different opinions on where `Generics` based code should live. Other than that, I found myself repeatedly making the same decisions that were made in Beam. Ultimately, I came to the realisation that there is not enough new in Rel8 to warrant yet another database library, so I have decided to put it on the shelf and call it a successful learning project, but to otherwise focus my attention on Beam.

I'm new to Beam, but not new to Haskell. Despite that, https://hackage.haskell.org/package/beam-core-0.6.0.0/docs/Database-Beam-Query.html#g:7 is outright terrifying. I spent a bit of time thinking, and I realised that this project could be a perfect fit for Backpack. Ordinarily, I'd open a conversation up at this point, but I wanted to put my money where my mouth is, and decided to try and run with that. This branch is the result of a few days of hacking, and I think the result is definitely something worth talking about.

My main observation was that `beam-core` is a database-agnostic library, and `beam-postgres` etc provide instantiations of the type classes to give you concrete queries you can run. The main thing that a backend really defines is the syntactic translation from beam to SQL, so the majority of my work was to rip out the `IsSql92*` type classes entirely, and instead build a signature of SQL syntax. This is `Database.Beam.Backend` in the `beam-backend` library. The next step is to go through all of `beam-core` and change any code that uses type classes to instead just use functions from this signature. The result is a considerably cleaner API. I've uploaded the Haddocks to my branch of `beam-core` here: https://ocharles.org.uk/beam-core/Database-Beam-Query.html#g:7 - look at those lovely types!

As I was learning the internals of `beam` at the same time, I might have made a few other changes. The point of this pull request is not to deliver you a finished branch, but to open up discussion, but also to give you something to at least try out to get a feel for my intention. The only "problematic" change I had to make was dropping `OrderingSyntax` and replacing it with `ExpressionSyntax`. I haven't ran this all the way through so I might have to drop back on that, but by making that change `QGenExpr` can just hold `WithExprContext ExpressionSyntax`. The alternative I'm considering is to have `QGenExpr` store syntax based on the context it's parameterised with, but this is going to require some type magic.

With that out the way, my thoughts on the user experience are that this is mostly an improvement:

* Documentation is considerably easier
* Type inference results in equally small types, rather than propagating all the syntactic requirements from the finally tagless encoding
* You can write database agnostic code still by just depending on `beam-core` and not mixing in a database backend.
* We can still differentiate between SQL92 and SQL2003, but we'll need a few more indefinite libraries for this. E.g., `beam-sql92` and `beam-sql2003`. `beam-postgres` would be able to instantiate both of these libraries.

The biggest drawback I can see is that it might no longer be possible to run *the same* query over two different database backends in the same project. This is because when you write a query, you are writing it against a specific `Q`, and `Q` ultimately mentions syntax. Backends define their own syntax, so you can't write a query once and run it in different backends. You can use multiple backends in a project, but this is going to require choosing backends at the time you write the query - I can expand on this if it's confusing.

If we really want to support this, I suggest actually having `beam-core` move back to a final deep embedding, and then having an indefinite library that takes this deep embedding and turns it into backend specific syntax. This ultimately might be cleaner, but it's a bigger rewrite. I'm happy to sketch this proposal out.

The finally tagless style works well when you are doing things like nano-passes over subsets of the grammar, or you want multiple interpreters. I don't really see either of these in beam. There is the reflected deep-embedding AST, and I've had to drop that. That does mean `dumpSQL` can no longer be used, but backends could export their own dumping routine.

Finally, please feel free to reply with no to this proposal! It's your project, and this is a rather anti-social way to propose a change, but it's the only way I could see to meaningfully proceed. I do hope we can find *something* that works to reduce the verbosity of these types though.


tathougies:
I will respond in more depth after I look into this more and try to understand how everything fits together, but just wanted to acknowledge this, and thank you for the in-depth proposal and implementation. It's always nice to try new ideas.

That being said, my main focus right now is getting everything on hackage/stackage, so no more major changes, until that's done 😄

ocharles:
I found some more time to chip away at this, and think I'm now happy with `beam-core`. I'm putting together https://github.com/ocharles/beam/blob/0f570a0b3603e63d5b2776d6bfef3326159ceb01/Changes.md to guide you through this PR. Please take your time though, there's no rush on my side - just trying to pre-empt whatever I can to make this easier.

3noch:
As a serious user of Beam I just want to voice a good measure of skepticism that we are *ready* for this. Backpack is so young. Do we know that losing the ability to write the same query against different backends is an acceptable loss? I know that I've wanted to leverage that before (but didn't get the chance). Does it work with GHCJS (several Beam projects use GHCJS with Beam types). Are there *other* ways of simplifying the types without resorting to Backpack?

I honestly don't know enough about backpack or its support in the ecosystem to actually answer these questions. But I'm definitely leery at this point and want to just advise caution.

That said, I definitely can sympathize with the desire for simpler types. They are truly terrifying at times. I've learned to work with them over time (lots of type aliases or lots of partial type signatures). But your work in that direction, @ocharles, is certainly appreciated. Also, I can imagine that Backpack could improve performance, albeit query generation is not the bottleneck in DB interactions by a long shot.

ocharles:
Scepticism is healthy, nothing wrong with that! As long as we're rational with our decision, I wouldn't want to base it solely on the fear of Backpack being new .That said, maybe I can rope @ezyang in here to comment on stability of Backpack. Are big changes on the horizon? I see now with things like https://hackage.haskell.org/package/unpacked-containers Backpack is starting to hit Hackage.

> Do we know that losing the ability to write the same query against different backends is an acceptable loss?

As you point out in [1], you only lose the ability to run the same query in multiple backends in the same package component. And even then, I think you can get around that with `mixin` magic, essentially duplicating a query module:

```
mixins: 
  renaming (MyApp.Query as MyApp.Query.SQLite) requiring (Database.Beam.Syntax as Database.Beam.SQLite.Syntax),
  renaming (MyApp.Query as MyApp.Query.PostgreSQL) requiring (Database.Beam.Syntax as Database.Beam.PostgreSQL.Syntax)
```

or something.

> Does it work with GHCJS (several Beam projects use GHCJS with Beam types).

I believe it should, but I can't say for sure. I imagine in this scenario, we might need a `beam-noop` or something library that fulfills the `Database.Beam.Syntax` signature, or you won't be able to link. That, or `beam-schema` gets broken out, which has no requirement on `Database.Beam.Syntax`.

> Are there other ways of simplifying the types without resorting to Backpack?

I'm not sure! Part of the reason I've just jumped right in to this is because I'm unable to answer the question "does Backpack help?" without just cracking on with it and seeing. Right now, pieces are falling into place remarkably well.

@3noch Would you be up for testing this when I've ironed everything out? If so, what libraries do you need? `beam-core`, `beam-postgres` and `beam-migrate`?

ezyang:
For prod users, I think the biggest show stopper for Backpack is lack of Stack support. You can use Backpack in Stack, but it's much more limited (it has to be done completely internally in convenience libraries; the public library isn't allowed to have signatures.) I am ostensibly working on adding support to this to Stack, but this requires "deep code diving into unfamiliar codebase time", of which I am sadly short of these days.

I don't think the lack of dynamic switching is a big loss. In particular, you can still write a typeclass interface / dynamic record on top of Backpack modules to get back dynamic switching in the cases you need it.

3noch:
@ocharles I'd certainly be willing to take a crack at building my app with your fork when it's ready. Unfortunately the app is closed source, so I can only offer high-level feedback. It's using `beam-core` and `beam-postgres`, but not `beam-migrate` (since it wasn't quite ready at the time the app was built).

> I don't think the lack of dynamic switching is a big loss. In particular, you can still write a typeclass interface / dynamic record on top of Backpack modules to get back dynamic switching in the cases you need it.

@ezyang, this is a good point.

Regarding stack, does it support public libraries that rely on signatures as dependencies but only expose a traditional Haskell API? If so, recreating Beam's current typeclass approach as a package could be a temporary solution for stack users.

tathougies:
Just as a point of fact, I do tend to use dynamic switching when I write tools using beam, so anything that loses this is a major pain point. One thing I actually like is having proper backend constraints so I know if a query won't work across a backend I may want to offer. Any solution that I accept is going to have to have a way to do this, even if it as a separate module.

ocharles:
> Just as a point of fact, I do tend to use dynamic switching when I write tools using beam, so anything that loses this is a major pain point.

I don't quite follow this. Do you mean you write a single executable that links in both `beam-sqlite` and `beam-postgres`, and allows the user to supply which type of DB they want to run a query against?

>  One thing I actually like is having proper backend constraints so I know if a query won't work across a backend I may want to offer.

With the current approach I'm exploring, sql99 and sql2003 are in different Hackage modules. If your library uses both of those, then you will have to compile it against a backend that has an implementation for both of those modules. `beam-core` is just sql93.

tathougies:
> I don't quite follow this. Do you mean you write a single executable that links in both beam-sqlite and beam-postgres, and allows the user to supply which type of DB they want to run a query against?

Yes, the Beam URI openers module makes this kind of stuff quite straightforward. Best of all, I get to write constraints against my backends, so if i want to introduce a new backend, I can pretty much expect it to work, given it matches the constraints. With this backpack approach there is no way to get this. I feel this is an important use case

An alternate approach to all of this is a much simpler one. We can keep `Database.Beam.Query`, scary types and all, except we won't export it from `Database.Beam`. We'll export it instead from `Database.Beam.Agnostic`. Then, we'll create modules in each backend `Database.Beam.Postgres.Prelude` and `Database.Beam.Sqlite.Prelude` that just re-exports everything. This has all the effect of making the haddocks nice, but we don't need to use any new features. Backpack is nice, but honestly, the syntax type variables are mainly a crutch. They're not super interesting for type-checking, but I don't see how to get rid of them (although, I'm going to try some experiments of my own later this week).

tathougies:
I've taken my own stab at improving the types. I introduce one new type class for backends, that subsumes all of the checks we used to do at the function level.

The signature for `all_`:

http://travis.athougies.net/haddocks/beam-core-0.8.0.0/Database-Beam-Query.html#v:all_

ocharles:
I've been on holiday for a while but just got back, and unfortunately have ran out of steam on this (as I kind of expected might happen). It's been a fun experiment, and I still remain convinced there is potential - but I got stuck on getting the changes added to `beam-migrate`. Maybe I'll look at this again in the future, but for now I'm going to abandon ship. Hopefully we can at least make smaller steps (like @tathougies's last comment) to improve the API.

PR #209, opened 2018-04-11
Added beam-discussion google link

stedi67:
I couldn't find any `beam-users` Google Group, but instead the `beam-discussion` one. This might help newcomers (as me) to end up in the right place.


tathougies:
You're right... can't believe I missed that :)

PR #220, opened 2018-04-21
auto-derive `primaryKey`

vwwv:
When the first rows from the table define the primary key, the function `primaryKey` can be derived.

If an embeded beam appears on the table holding some columns used for the primary key, then the whole embeded beam shall appear on the primary key as is. This restriction could be lifted, so it would be possible to use "half" beam on the key (or even more, use a different set of beams on the primary keys and table) by using `Generic1` or ` {-# OVERLAPS #-}`, but I'm not sure that would be a good idea (or is there some suitable generalization that is appropriate?)

Example:
```
{-# LANGUAGE DerivingStrategies, DeriveAnyClass #-}

import Database.Beam
import Data.Text (Text)


data UserT f
    = User
    { _userEmail     :: Columnar f Text
    , _userFirstName :: Columnar f Text
    , _userLastName  :: Columnar f Text
    , _userPassword  :: Columnar f Text }
    deriving stock    Generic
    deriving anyclass Beamable

type User = UserT Identity
type UserId = PrimaryKey UserT Identity
deriving instance Show User
deriving instance Show UserId
deriving instance Eq User


instance Table UserT where
    data PrimaryKey UserT f = UserId (Columnar f Text) deriving stock    Generic
                                                       deriving anyclass Beamable 

main :: IO ()
main = let user = UserT "email@server.com" "John" "Johnson" "*****"
        in print ( primaryKey user :: UserId )
```

### DIFF
--- a/beam-core/Database/Beam/Schema/Tables.hs
+++ b/beam-core/Database/Beam/Schema/Tables.hs
@@ -56,6 +56,7 @@ module Database.Beam.Schema.Tables
     where
 
 import           Database.Beam.Backend.Types
+import           Database.Beam.Schema.Tables.GPrimaryKeyOf(GPrimaryKeyOf,gPrimaryKey)
 
 import           Control.Arrow (first)
 import           Control.Monad.Identity
@@ -520,7 +521,15 @@ class (Typeable table, Beamable table, Beamable (PrimaryKey table)) => Table (ta
 
     -- | Given a table, this should return the PrimaryKey from the table. By keeping this polymorphic over column,
     --   we ensure that the primary key values come directly from the table (i.e., they can't be arbitrary constants)
+    --
+    --   Whenever `PrimaryKey table column` and `table column` are generic, and the first rows from `PrimaryKey table column`
+    --   match those from `table column`, this function can be auto derived.
     primaryKey :: table column -> PrimaryKey table column
+    default primaryKey :: ( Generic (table column)
+                          , Generic (PrimaryKey table column)
+                          , GPrimaryKeyOf (Rep (PrimaryKey table column)) (Rep (table column))
+                          ) => table column -> PrimaryKey table column
+    primaryKey = to . fst . gPrimaryKey . from 
 
 -- | Provides a number of introspection routines for the beam library. Allows us
 --   to "zip" tables with different column tags together. Always instantiate an
--- a/beam-core/Database/Beam/Schema/Tables.hs
+++ b/beam-core/Database/Beam/Schema/Tables.hs
@@ -56,7 +56,8 @@ module Database.Beam.Schema.Tables
     where
 
 import           Database.Beam.Backend.Types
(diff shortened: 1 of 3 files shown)


vwwv:
> ... you could use GHCs type errors to make it give a sensible error ...

Sure, nice suggestion, I'll add it. 
I think changing
```
instance GPrimaryKeyOf (K1 info1 a) (K1 info2 a) where
```
to something like
```
instance (a ~ a', ErrorIfNotTheSame a a') => GPrimaryKeyOf (K1 info1 a) (K1 info2 a') where
```
should be easily doable.

vwwv:
So now there's better error report :)

But I need to use `INCOHERENT` and `error` in a couple of places...

tathougies:
I don't think you need `INCOHERENT` here, if you combine both type families. The use of error is fine.

vwwv:
> I don't think you need `INCOHERENT` here, if you combine both type families.

I'm not sure I get you, do you mean the type families to report the error? (`ReportToLongPrimaryKey` and `ReportMissfitError`) ?

If so, the "incoherence" is actually between these instances:

```
instance {-# INCOHERENT #-} 
         ( ReportMissfitError debugInfo
         ) => GPrimaryKeyOf debugInfo (K1 info1 a) (K1 info2 b)

instance GPrimaryKeyOf debugInfo (K1 info1 a) (K1 info2 a) 
   where
    type GRemainds (K1 info1 a) (K1 info2 a) = U1
    gPrimaryKey _ (K1 x) = (K1 x, U1)
```

I've tried several ideas to avoid that problem, such using closed type families to return a different type whether `a ~ b` and then use that value to disambiguate the instances...but it didn't work because the type family wasn't getting applied (ghc was not able to tell the result). I think the reason is ghc is sometimes able to tell `a ~ b`, but It can't tell `a /~ b`.

As far as I'm aware, treating `a ~ b` different than `a /~ b` is incoherent in itself no matter how it is implemented. Though maybe I'm missing something or there's another approach, do you have some suggestions?

Another option if you don't like the `INCOHERENT` pragma, could be using avoiding the incoherent instance and add on the other one an explicit `~`, such `(a~a')=>GPrimaryKeyOf debugInfo (K1 info1 a) (K1 info2 a')`. This way, the error message becomes simpler, something like `Could not deduce Cf Int ~ C f Char`, which is worse than a custom error message, but at least is simple and somehow related.

vwwv:
Trying to solve the problem, the most I could was to defered the incoherent issue to the calling place, but then there, say `C f Int` can not be proved to be different to `C f Char` without knowing `f`. (or was `C` injective?)

tathougies:
Ah I see.. my main concern with incoherent instances is that they will cause type checking to fail for otherwise correct programs. I will take a more in-depth look at this PR sometime this weekend.

kmicklas:
Closing this due to inactivity. Feel free to reopen.

While auto-deriving things is always nice, personally, I think the complexity to usefulness ratio of this is pretty high. If we could find a way to derive based on the field name symbol rather than trying to match the types, I think that would be preferable.

PR #230, opened 2018-05-04
Add leftJoin_' example to the tutorial

mulderr:
I was trying to verify things by:
```
nix-shell
./build-docs.sh servedocs
```
But it takes forever to build. I see `tar Jxf ~/stack/programs/x86_64-linux-nix/ghc-8.2.2.tar.xz` being run repeatedly, almost all the time. That seems very, very unnecessary and is consuming lots of CPU time.

I'm not really familiar with mkdocs. Any pointers?


mulderr:
Aha, ghc setup fails with:
```
/bin/bash: bad interpreter: No such file or directory
```
Very true since this is NixOS. Now only need to figure out where to fix that. Could also make a symlink manually but that's just... nah.

tathougies:
Huh.... how strange.... it seems like stack is trying to unpack ghc? I think have stack set to use nixos’s ghc on my machine. Can you try setting that up?

3noch:
Yeah on NixOS you need to tell stack to use system GHC. You can use `/usr/bin/env bash` in NixOS.

mulderr:
Yep, that's it. I'm not sure where the interpreter error was coming from but it may not even be directly related. I just added:
```
nix:
  enable: true
```
to `~/.stack/config.yaml` and now it's working fine.

To be clear this wasn't a mkdocs problem. `stack setup` on a new project was failing as well. I don't use stack normally so I didn't have things pre-configured.

tathougies:
Oh... interesting. I'll add something to README. Is this good to go?

tathougies:
Great... thank you!

PR #233, opened 2018-05-07
derivable parametrized beams

vwwv:
Generalize the deriving mechanism so the class `Beamable` can be derived
even for those types containing parametrized `Beamable` fields.

For example, the following code does not currently compile, but it would under this PR:
```
module Foo where

import Database.Beam


data FooT inner f = Foo 
      { _foo  :: C f Int
      , _moo  :: C f Int
      , _inn1 :: inner f
      } deriving(Generic)

instance (Beamable inner) => Beamable (FooT inner)
```

This is a fix for issue https://github.com/tathougies/beam/issues/222

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -11,3 +11,4 @@ Beam would like to thank the following people for their contributions to beam.
 "Sajid Ibne Anower"<sajidanower23@gmail.com>
 "Brian Jones"<brian@uncannyworks.com>
 "Sam Protas"<sam.protas@gmail.com>
+"Alejandro Duran Pallares"<vwwv@correo.ugr.es>
--- a/beam-core/Database/Beam/Schema/Tables.hs
+++ b/beam-core/Database/Beam/Schema/Tables.hs
@@ -477,8 +477,15 @@ from' = from
 to' :: Generic x => Rep x () -> x
 to' = to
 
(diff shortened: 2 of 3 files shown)


vwwv:
@tathougies, would the tests from `Database.Beam.Test.Schema` still apply for this or would this require a different set of test? is so, any hint about how they should be?

tathougies:
The tests in `Database.Beam.Test.Schema` should still pass, of course, but if you could follow their format and write a test that demonstrates usage of a parameterized `Beamable`, that'd be great! The test doesn't need to do very much, but it should verify that the code compiles and that the defaults generated are reasonable.

Didn't take an in-depth look, but it looks good so far... great work!

vwwv:
Ok, I'll add then a test like that to this PR

tathougies:
This patch is looking really great! Just one thing. It looks like you test that it compiles with embeddable `Beamable`s, but we also want to make sure the names are correct, and follow some kind of pattern. There are some examples of these tests in the Test.hs file.

Add that, and I'll merge this. Thank you very much!

Also, please add yourself to the `CONTRIBUTORS` file in the top-level. Thanks again!

vwwv:
> Also, please add yourself to the CONTRIBUTORS

Great, thanks!

> ...make sure the names are correct, and follow some kind of pattern...

ok, I'll add it to the test.

vwwv:
I've added a couple of tests, let me know if it is nor ready to be merged or if it needs some change.

PR #240, opened 2018-05-21
Fix default SQL type for Double

mulderr:
Got some unexpected failed predicates from `verifySchema` where beam would insist it wants `Double` to be stored as `real`. [Docs](https://www.postgresql.org/docs/current/static/datatype-numeric.html) suggest this should be `double precision`.

Not quite sure how much breakage this could cause so just throwing it out here for now.


mulderr:
@tathougies Any thoughts on this one?

`real` and `double precision` are the same thing on sqlite but that is a sqlite exclusive weirdness afaik. They definitely have different sizes on postgres where having `Double` mapped to `real` effectively truncates values to single precision.

Should `autoMigrate` be able to figure out that it can change the type from `real` to `double precision` without loosing data? If not then a version bump and a warning in the change log may be in order because `verifySchema` and `autoMigrate` among others may start failing after an upgrade and some manual `alter table`s may be necessary.

I am in the comfortable situation of not having anything in production relying on migrations yet (but I'd like to, soon!) so I wouldn't want to force a breaking change on others.

If anyone has any feedback or can confirm either way that would be great.

tathougies:
I think you're right that Double should be a double and not a real. I'm not sure why I had that to begin with. I think this should go along with a minor version bump. I plan on making some large-ish changes soon, so that's fine.

PR #241, opened 2018-05-22
Ignore dropped columns in getDbConstraints

mulderr:
`getDbConstraints` must filter out dropped columns when gathering info for column checks. Otherwise the query could result in something like below. Tested on 9.6.8.

```
           attname            | atttypid | atttypmod | attnotnull |         format_type
------------------------------+----------+-----------+------------+-----------------------------
 id                           |       20 |        -1 | t          | bigint
 (...)
 ........pg.dropped.7........ |        0 |        -1 | f          | -
 ........pg.dropped.8........ |        0 |        -1 | f          | -
 ........pg.dropped.9........ |        0 |        -1 | f          | -
```


mulderr:
Btw. after #240 and this I finally got `autoMigration` working.

tathougies:
Nice catch!

PR #248, opened 2018-05-30
Postgres Range type support

SamProtas:
I noticed this appeared missing from the Postgres specific package.

I admittedly haven't tested this locally (yet) so there could be syntax bugs, missing constraints, etc. but I figured I'd gone far enough without some feedback (maybe too far?).

I'm also pretty sure this doesn't support beam-migrate yet. I'm not familiar with that functionality at all but I'd be willing to add that support to this PR with a bit of guidance.

When you get a chance, let me know what you think of the general design or any questions/comments you have.

Next chance I get to work on this I'll confirm some basic schema definitions and queries, assuming you don't have any fundamental design change requests to take care of first.

### DIFF
--- a/beam-postgres/Database/Beam/Postgres.hs
+++ b/beam-postgres/Database/Beam/Postgres.hs
@@ -7,7 +7,7 @@
 --
 -- @beam-postgres@ supports most beam features as well as many postgres-specific
 -- features. For example, @beam-postgres@ provides support for full-text search,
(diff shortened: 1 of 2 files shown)


SamProtas:
FYI I've confirmed things compile and the syntax is correct for many of the functions/operators. I spot-checked those that had different implementations so I'm fairly confident the rest are okay as well.

tathougies:
Cool!

tathougies:
I'm happy merging this.. Thanks for all the work!

Issue #258, opened 2018-06-15
Documentation: Autoincrement problem in part 2 of tutorial


sevanspowell:
I believe the auto increment keyword used in this part of tutorial 2:

```
sqlite> CREATE TABLE addresses ( id INT AUTO_INCREMENT PRIMARY KEY, ...
```

should be:

```
sqlite> CREATE TABLE addresses ( id INTEGER PRIMARY KEY AUTOINCREMENT, ...
```

Or something similar, otherwise I don't get an auto-incrementing id column in the address table.

sevanspowell:
Submitted PR #259.

sevanspowell:
Already fixed with 4b1fdd7.

PR #259, opened 2018-06-15
Fix Address table creation statement in part 2 of the tutorial

sevanspowell:
- Fixed misspelled 'AUTO_INCREMENT' -> 'AUTOINCREMENT'.
- Used INTEGER instead of INT to allow the above change and to match the
  creation statement used in part 3 of the tutorial.
- Addresses issue #258


tathougies:
Ah I see, the issue with the CI is that you want to push to docs master. Can you re-open the PR but pushing to master instead? I'll push to docs-master later.

sevanspowell:
Thanks @tathougies, looks like this has already been fixed on master (4b1fdd7), so I'll close this PR.

PR #261, opened 2018-06-16
Tutorial 1 fixes

timhabermaas:
I fixed some small errors when going through the first tutorial.

Adding `MultiParamTypeClasses` in GHCI is only necessary when following the tutorial in GHCI past the "Defining our database" headline. However, I doubt anyone is still using GHCI at this point, because you can't c&p multiline definitions into GHCI, so many will probably already give up when trying to define `UserT` (at least I did).

Maybe it would be easier to follow when telling people to create a file (`Main.hs`) and load that into GHCI (`:l Main` / `:r` + functions `insertUsers`, `query1`, ...)?

### DIFF
--- a/docs/tutorials/tutorial1.md
+++ b/docs/tutorials/tutorial1.md
@@ -111,7 +111,7 @@ values of type `User`. We can use the `StandaloneDeriving` and
 'regular' datatype.
 
 ```
-> :set -XStandaloneDeriving -XTypeSynonymInstances
+> :set -XStandaloneDeriving -XTypeSynonymInstances -XMultiParamTypeClasses
 ```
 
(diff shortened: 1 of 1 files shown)


timhabermaas:
I just learned about `:{`, `}:`: https://stackoverflow.com/a/8443096 So, multiline definitions are actually possible in GHCI.

I'd still prefer a literal Haskell-style tutorial because it's easier to play around with the code.

tathougies:
Why do you need the `text` package?

timhabermaas:
> Why do you need the `text` package?

I need it for the import of `Data.Text` in

```haskell
import Database.Beam
import Database.Beam.Sqlite

import Data.Text (Text)
```

PR #266, opened 2018-06-27
Fix stack command missing dependencies for in the repl

jappeace:
https://github.com/tathougies/beam/issues/264


tathougies:
Duplicate of #261

PR #283, opened 2018-07-23
Adds instances for UTCTime in the Postgres backend

### LINKED ISSUE #282: add the instance for HasDefaultSqlDataType PgDataTypeSyntax UTCTime
This looks like an omission, since the corresponding HasSqlValueSyntax instance is provided

```haskell
instance HasSqlValueSyntax PgValueSyntax UTCTime where
  sqlValueSyntax  = defaultPgValueSynatx
```
(using CPP in [Database/Beam/Postgres/Syntax.hs](https://github.com/tathougies/beam/blob/b88ef2ed69815c8440bfdf80afd539055e92e219/beam-postgres/Database/Beam/Postgres/Syntax.hs#L1112))

adetokunbo:
Assumes that the instance defintions for LocalTime should be OK for UTCTime,
although perhaps the LocalTime HasSqlDataType definition ought to be changed to

```haskell
    defaultSqlDataType _ _ timestampTyep Nothing True
```

since LocalTime is expected to reflect the timezone of the database ?

Fixes #282

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Syntax.hs
+++ b/beam-postgres/Database/Beam/Postgres/Syntax.hs
@@ -1328,6 +1328,10 @@ instance HasDefaultSqlDataType PgDataTypeSyntax LocalTime where
   defaultSqlDataType _ _ = timestampType Nothing False
 instance HasDefaultSqlDataTypeConstraints PgColumnSchemaSyntax LocalTime
 
+instance HasDefaultSqlDataType PgDataTypeSyntax UTCTime where
+  defaultSqlDataType _ _ = timestampType Nothing False
+instance HasDefaultSqlDataTypeConstraints PgColumnSchemaSyntax UTCTime
+
 instance HasDefaultSqlDataType PgDataTypeSyntax (SqlSerial Int) where
   defaultSqlDataType _ False = pgSerialType
   defaultSqlDataType _ _ = intType


tathougies:
`UTCTime` also reflects the timezone of the database, which is the main reason it's not here. Postgres does not do timestamp conversions to UTC. I'm not sure why everyone thinks it does.

adetokunbo:
I'm not sure I understand when you say UTCTime also reflects the timezone of the database. There is seems to be a distinction between timestamp values declared WITH TIMEZONE and WITHOUT.  I have assumed that this distinction can be mapped to the difference between UTCTime and LocalTime as follows:

From: [postgress datatype-datetime](https://www.postgresql.org/docs/9.1/static/datatype-datetime.html)

> In a literal that has been determined to be timestamp without time zone, PostgreSQL will silently ignore any time zone indication. That is, the resulting value is derived from the date/time fields in the input value, and is not adjusted for time zone.

- This seems to correspond to the UTCTime type ?

> For timestamp with time zone, the internally stored value is always in UTC (Universal Coordinated Time, traditionally known as Greenwich Mean Time, GMT). An input value that has an explicit time zone specified is converted to UTC using the appropriate offset for that time zone. If no time zone is stated in the input string, then it is assumed to be in the time zone indicated by the system's timezone parameter, and is converted to UTC using the offset for the timezone zone.

- this seems to corresponds to something declared as a LocalTime.  Even though the actual value may be provided (by Beam to the database) without an explicit timezone, the database interprets this as time in UTC and henceforth keeps track of that if the timezone of the database changes.

Please let me know what's wrong this approach

mulderr:
> this distinction can be mapped to the difference between UTCTime and LocalTime as follows: ...

I'd say that's backwards.

UTCTime is a special case of a timestamp with time zone. The time zone is simply always UTC.

If you give me a timestamp and don't state the time zone, it's not really sensible to assume UTC. One should probably not represent it as UTCTime then. On the other hand LocalTime makes sense. It is a local time *somewhere*.

adetokunbo:
> If you give me a timestamp and don't state the time zone, it's not really sensible to assume UTC. One should probably not represent it as UTCTime then. On the other hand LocalTime makes sense. It is a local time somewhere.

I would have thought the answer to that depends on implementation choices. 
E.g, if, as part of a system's design, it never stores timestamps in anything other than UTC, then by definition its always sensible to assume UTC and represent it as UTCTime.  This is a fairly reasonable implementation choice; according to the documentation, this is what Postgres itself does.

mulderr:
UTCTime from timestamp without time zone comes with additional assumptions you cannot enforce. To me, if read literally, this instance would state: you can get a UTCTime from *any* timestamp without time zone. Even from say an existing database. At the very least this could lead to differences in interpretation.

On the other hand I see your point. It's just practical to have it anyway. And there already is one in [beam-sqlite](https://github.com/tathougies/beam/commit/5c8d96ae9a9f930a2f860bf282fb6954ce697bcf).

> Postgres does not do timestamp conversions to UTC. I'm not sure why everyone thinks it does.

Not for timestamp without time zone. It surely does for timestamp with time zone - as cited above by @adetokunbo. Also:

> When a timestamp with time zone value is output, it is always converted from UTC to the current timezone zone, and displayed as local time in that zone. To see the time in another time zone, either change timezone or use the AT TIME ZONE construct 

I wasn't aware timestamps with and without time zone are stored exactly the same. They're just interpreted differently.

The ZonedTime instance must necessarily be a bit weird then. You may not expect any conversions and be in for a nasty surprise when you read things back.

Btw. to make things easier different database engines do it differently. In Oracle timestamp with time zone does store the time zone and there is additionally timestamp with local time zone that is more like timestamp with time zone in postgres except stored in database time and not utc.

3noch:
IMO `UTCTime` should be a particular decoding of PG's `TIMESTAMP WITH TIME ZONE`. If you have a time zone, you can always get back to UTC safely. If you don't have a time zone at all, then you presumably must look elsewhere (perhaps another column, table, setting file, etc.) to know how to decode it. This is outside the scope of Beam's instances. Thus `LocalTime` should be used for anything at all that comes from the database without a time zone. There is still work left to be done do convert that to a `UTCTime` (even if that work is trivially "assert UTC"). If the column has time zone information then it could be decoded either as `ZonedTime` or `UTCTime`, where the latter applies the offset and throws it away because you don't care about it.

3noch:
Hm, but if I'm understanding correctly, PG itself does not store time zone information but only uses it to adjust from/to UTC. If this is the case then `UTCTime` instance should tell PG provide the timestamp value in UTC.

tathougies:
It may be possible to ask postgres to convert a timestamp to UTC. I'll read the docs and see what I can come up with.

3noch:
`AT TIME ZONE 'UTC'`

3noch:
@tathougies Would it be possible to add an instance for `UTCTime` using `AT TIME ZONE 'UTC'`?

tathougies:
@3noch I'm really hoping so, but it would require changes to query generation that take into account the type of the expected output. Currently, beam does not do that. I'm working on this in the beam 0.8.0.0 branch (not pushed yet).

3noch:
@tathougies Excellent. Yeah I wondered if something like that would need change. Thanks for working on it.

tathougies:
Closing this in favor of implementing my fix (coming shortly...)

PR #287, opened 2018-08-01
Bumps upper bounds on Free

jkachmar:
Bumps the upper bounds on Free, which should (hopefully) unblock the `beam-*` family from Stackage snapshots.

I tested locally with the latest nightly snapshot and it all seems to work


tathougies:
Thank you!

PR #288, opened 2018-08-04
Convert FromBackendRowM into a free applicative

mulderr:
This is a work in progress.

There are already some clear trade-offs to this approach so I'm submitting here for review.

Pros:
* Result can be read sequentially i.e. no backtracking/lookahead. With a little luck we can implement `ParseAlternative` so that each column in the result is read exactly once. `CheckNNull` is thus gone - it would be useless with only applicative anyway.
* Every `fromBackendRow` can be statically analyzed to get the [number of values needed](https://github.com/mulderr/beam/blob/3c6e9e679637002745774cce9e988635cea8efeb/beam-core/Database/Beam/Backend/Types.hs#L78) without having to define it at the instance level.
* Backend libraries can do their own weird things like [inspecting types of fields](https://github.com/mulderr/beam-oracle/blob/fab1f03fb1cd85e7a41eff8a191e4b94fdac293f/src/Database/Beam/Oracle/Connection.hs#L78) without having to impose new functionality on beam-core.

Cons:
* Errors regarding nullability can be delayed and less informative, we basically thread null up the chain until a `Maybe` "catches" it or the whole computation results in null which should cause the interpreter to throw. The problem is exacerbated by nested `Maybe`s resulting from outer joins - otherwise we could possibly short circuit in the `Applicative` instance of `Result` by having a mix of `Null` and `Result x` be an `Error`.
* Well, to point out the elephant in the room, applicative is more restrictive than monad. Currently I cannot see any instances that strictly require monad but that doesn't mean anything. If you can think of an immediately useful one (or one that requires more than `FromBackendRowF` as defined here permits) do say!

TODO:
- [ ] try to improve error reporting, it's not been given much care up until this point
- [x] `Database.Beam.Postgres.Conduit` still has some things commented out
- [ ] fix errors when building docs for beam-postgres
- [ ] benchmarking

### DIFF
--- a/beam-core/Database/Beam/Backend/Types.hs
+++ b/beam-core/Database/Beam/Backend/Types.hs
@@ -6,7 +6,7 @@
 
 module Database.Beam.Backend.Types
   ( BeamBackend(..)
-  , Result (..), BeamRowError (..)
+  , Result (..)
 
   , FromBackendRowF(..), FromBackendRowA
   , parseOneField, parseAlternative
@@ -18,10 +18,10 @@ module Database.Beam.Backend.Types
 
 import           Control.Applicative
 import           Control.Applicative.Free
-import           Control.Exception
 import           Control.Monad.Identity
 import           Data.Monoid (Sum(..))
 import           Data.Tagged
+import           Data.Text (Text)
 import           Data.Vector.Sized (Vector)
 import qualified Data.Vector.Sized as Vector
 
@@ -31,16 +31,10 @@ import GHC.Generics
 import GHC.TypeLits
 import GHC.Types
 
-data BeamRowError
-  = BeamRowError String
-  deriving (Eq, Show)
-
-instance Exception BeamRowError
-
 data Result a
   = Result a
   | Null
-  | Error BeamRowError
+  | Error Text
   deriving (Eq, Show, Functor)
 
 instance Applicative Result where
--- a/beam-core/Database/Beam/Backend/Types.hs
+++ b/beam-core/Database/Beam/Backend/Types.hs
@@ -7,7 +7,7 @@ module Database.Beam.Backend.Types
   ( BeamBackend(..)
 
   , FromBackendRowF(..), FromBackendRowA
-  , parseOneField, peekField
+  , parseOneField, parseAlternative
 
   , FromBackendRow(..)
   , valuesNeeded
@@ -34,15 +34,17 @@ class BeamBackend be where
 
 data FromBackendRowF be f where
   ParseOneField :: BackendFromField be a => (Maybe a -> f) -> FromBackendRowF be f
-  PeekField :: BackendFromField be a => (Maybe a -> f) -> FromBackendRowF be f
+  ParseAlternative :: (BackendFromField be a, BackendFromField be b)
+                   => (a -> Maybe r) -> (b -> Maybe r) -> (Maybe r -> f) -> FromBackendRowF be f
 deriving instance Functor (FromBackendRowF be)
 type FromBackendRowA be = Ap (FromBackendRowF be)
 
 parseOneField :: BackendFromField be a => FromBackendRowA be (Maybe a)
 parseOneField = liftAp (ParseOneField id)
 
-peekField :: BackendFromField be a => FromBackendRowA be (Maybe a)
-peekField = liftAp (PeekField id)
+parseAlternative :: (BackendFromField be a, BackendFromField be b)
+  => (a -> Maybe r) -> (b -> Maybe r) -> FromBackendRowA be (Maybe r)
+parseAlternative f g = liftAp (ParseAlternative f g id)
 
 class BeamBackend be => FromBackendRow be a where
   fromBackendRow :: FromBackendRowA be (Maybe a)
--- a/beam-core/Database/Beam/Backend/Types.hs
+++ b/beam-core/Database/Beam/Backend/Types.hs
@@ -55,12 +55,13 @@ class BeamBackend be => FromBackendRow be a where
   default fromBackendRow :: BackendFromField be a => FromBackendRowM be a
   fromBackendRow = parseOneField
 
-  valuesNeeded :: Proxy be -> Proxy a -> Int
-  valuesNeeded _ _ = 1
-
   rowRep :: Proxy be -> Proxy a -> [AnyField be]
   default rowRep :: BackendFromField be a => Proxy be -> Proxy a -> [AnyField be]
   rowRep _ _ = [AF (Proxy @a)]
+
+valuesNeeded :: FromBackendRow be a => Proxy be -> Proxy a -> Int
+valuesNeeded be a = length $ rowRep be a
+
 
 -- | newtype mainly used to inspect tho tag structure of a particular
 --   'Beamable'. Prevents overlapping instances in some case. Usually not used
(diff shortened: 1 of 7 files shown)


mulderr:
Interesting. Did some basic benchmarks on postgres today and (seemingly) found a big regression:

Test query:
```haskell
-- doesn't make sense, want any ~200k rows of Maybe Text
large200kmt :: Pg [Maybe Text]
large200kmt = runSelectReturningList $ select $ do
  t1 <- all_ $ customer chinookDb
  t2 <- all_ $ track chinookDb
  pure (customerCompany t1)
```

master:
```
benchmarking 200k/maybe-text
time                 117.5 ms   (113.9 ms .. 122.2 ms)
                     0.999 R²   (0.997 R² .. 1.000 R²)
mean                 119.1 ms   (117.8 ms .. 120.3 ms)
std dev              1.818 ms   (1.208 ms .. 2.525 ms)
variance introduced by outliers: 11% (moderately inflated)
```

free-ap:
```
benchmarking 200k/maybe-text
time                 314.1 ms   (311.2 ms .. 316.3 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 312.9 ms   (311.5 ms .. 314.2 ms)
std dev              1.501 ms   (1.227 ms .. 1.659 ms)
variance introduced by outliers: 16% (moderately inflated)
```

Almost 3 times slower. Ouch!

Except, after a closer look... we're comparing very different things. On master it stops after the first Nothing and the length of the result is 1 - it so happens the result begins with a Just followed by a Nothing. On free-ap it returns the full 206677 rows - same as a raw query. I seem to recall Nothing is also being used to indicate "no more rows" in at least one place.

@tathougies What is the expected behavior here?

mulderr:
Aha! Right [here](https://github.com/tathougies/beam/blob/d730e26f52a196fd1e667e9ab1d675514c1b8672/beam-postgres/Database/Beam/Postgres/Connection.hs#L269) if I change `next r Nothing` to `next (Just r) Nothing` it seems to fix things and the query returns a all the rows.

tathougies:
@mulderr are you saying if you make that change, the current master produces all 206677 rows? If so, then this is likely a bug, and ought to be fixed regardless. What is the benchmark comparison like now?

tathougies:
the expected behavior is that all rows returned by the database ought to be surfaced to the application. If the expected row datatype is `Maybe a` and postgres returns 206677 rows of `Nothing`, we should receive all 206677 rows.

mulderr:
Yes, with that one change it returns exactly the expected number of rows.

It's kind of interesting how it all fits together. I think it was *always* parsing a `Maybe`. A null would then produce `Nothing` and cause the default implementation of `runReturningList` to interpret that as end of data.

There is only a minimal difference in the benchmarks now.

master (with fix):
```
benchmarking 200k/maybe-text
time                 312.3 ms   (286.8 ms .. 334.7 ms)
                     0.998 R²   (0.996 R² .. 1.000 R²)
mean                 289.4 ms   (276.8 ms .. 300.7 ms)
std dev              13.17 ms   (10.80 ms .. 14.92 ms)
variance introduced by outliers: 16% (moderately inflated)
```

free-ap:
```
benchmarking 200k/maybe-text
time                 330.0 ms   (312.8 ms .. 346.5 ms)
                     0.998 R²   (0.994 R² .. 1.000 R²)
mean                 319.9 ms   (309.6 ms .. 327.3 ms)
std dev              10.09 ms   (3.402 ms .. 13.60 ms)
variance introduced by outliers: 16% (moderately inflated)
```

mulderr:
Created #291 in case the fix looks good.

beam-core/Database/Beam/Backend/Types.hs
```diff
@@ -25,34 +30,54 @@ import GHC.Generics
 import GHC.TypeLits
 import GHC.Types
 
+data BeamRowError
```
tathougies: use a newtype or get rid of this error. Does the Exception instance imply this type will be thrown?

beam-postgres/Database/Beam/Postgres/Conduit.hs
```diff
@@ -147,10 +147,10 @@ runQueryReturning conn x withSrc = do
               do fields' <- liftIO (maybe (getFields row) pure fields)
                  parsedRow <- liftIO (runPgRowReader conn 0 row fields' fromBackendRow)
                  case parsedRow of
-                   Left err -> liftIO (bailEarly row ("Could not read row: " <> show err))
-                   Right parsedRow' ->
-                     do C.yield parsedRow'
-                        streamResults (Just fields')
+                    Left err -> liftIO (bailEarly row ("Could not read row: " <> show err))
```
tathougies: No need for the indentation here

mulderr:
It's extremely hard to get consistent results but my very crude benchmarks indicate up to ~20% slowdown in places I didn't expect. One of those is parsing rows of `Text`. On the other hand parsing `Int`s is sometimes up to ~20% faster. So, it looks like `ParseOneField` is slower in general but `fromPgIntegral` is possibly much faster and makes up for it? This could suggest that `fromPgIntegral` in master can be improved. Anyway I'd like to see what can be done about `Text` and where the slowdown comes from.

I've also experimented with `Control.Applicative.Free.Final` and `Control.Applicative.Free.Fast`. At first glance they don't seem to be clearly better or worse at least in the tests I'm doing. You can easily swap those - just replace the imports.

mulderr:
My conclusion so far is that `Ap` can be marginally faster for privimite results (could just be minor tweaks and inlining) but overall complexity is worse and for complex (nested?) result types the differences are big.

GHC profiling output shows most time (~25%) is spent in fmap (NB: `<*>` for `Ap` is implemented in terms of fmap).

In hindsight this probably makes perfect sense. We are comparing a regular ADT representation of `Ap` with Church encoded `F`.

So while `Ap` can definitely work and would provide more flexibility IMHO the performance cost is very much not acceptable and I cannot see a magic way to improve it. One last thing before I call it - I wonder if it's possible to have `Ap` on the outside and transform it to `F` or another representation for reading.

tathougies:
Thanks for all the work. Let me play around with it this weekend. There may be some ways to get the time down. One thing to think about more is inlining. You say it is slightly faster for primitive types and the slow down is around complex types, but one possibility here is that inlining is not firing where it should. If we could get that to work, then it could be faster. For the most part, GHC ought to know the 'shape' of the data at compile time.

mulderr:
I have tried a couple different things but every time I'm seeing lots of time being spent in `<$>` or `<*>` of FromBackendRow. Even after converting `Ap` to `F` for reading rows it's still not gone which makes me think either I'm not fully converting things (?) or somehow laziness gets the last laugh.

Anyway to me the problem revolves around providing efficient Maybe and Generic instances.

First, I'm pretty convinced that the current handling of Maybe is already surprisingly efficient. GHC knowing the shape of things doesn't help at all here. If we avoid checking null for each sub part ahead of time then we have to create and then `<*>` a bunch of `Maybe`s. So it makes me think the current approach may actually be a really smart thing to do.

Also, I cannot grasp why [this fmap](https://github.com/tathougies/beam/blob/master/beam-core/Database/Beam/Backend/Types.hs#L150) basically disappears for `F` and not for any version of `Ap` I tried. It may be the place to look for more specialization/inlining as you say.

Finally, wrapping things in a `Result` or so is not a good idea as more nesting causes even more allocation and even more processing. Any formulation of `FromBackendRow` using `Ap` that wants to compete with `F` would have to find a way to avoid that and probably embed facilities for handling `Maybe` directly in `FromBackendRowF`.

kmicklas:
Feel free to reopen and revive this if conflicts and the performance regressions are fixed, but I'm going to close for now due to inactivity.

PR #289, opened 2018-08-06
Bumps remaining free bounds

jkachmar:
Looks like I missed two modules the first time around 😅, this should hopefully be all that's needed to get `beam-*` into Stackage Nightly.

I also added `stack.yaml` files to `beam-postgres` and `beam-sqlite` with similar contents to the ones in other modules. I can remove these if they were left out deliberately.

### DIFF
--- a/beam-postgres/beam-postgres.cabal
+++ b/beam-postgres/beam-postgres.cabal
@@ -38,7 +38,7 @@ library
 
                       hashable             >=1.1  && <1.3,
                       lifted-base          >=0.2  && <0.3,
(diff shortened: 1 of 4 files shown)


tathougies:
Stack.yaml in beam-postgres and beam-sqlite were left out deliberately. If you only want to build one or the other, just run `stack build beam-postgres` or `stack build beam-sqlite` in the top-level directory. I'm happy to merge the PR after this.

jkachmar:
Removed the files 👍

PR #291, opened 2018-08-08
Fix stepProcess

mulderr:
Fix bug in withPgDebug's stepProcess.

In case of a successful parse we should always pass a Just to next.


tathougies:
Thank you!

tathougies:
Thanks!

PR #292, opened 2018-08-10
Fix build on ghc-8.4.3

ryantrinkle:
* I haven't tested in this on other GHC versions, but I did try to avoid breaking them
* Is there a way I can/should add GHC-8.4.3 to a test suite and/or CI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,10 @@ jobs:
       script: ./ci/test
       env: BEAM_RESOLVER=lts-12.0
 
+    - stage: test
+      script: ./ci/test
+      env: BEAM_RESOLVER=lts-12.5
+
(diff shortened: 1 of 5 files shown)


tathougies:
You can add a new stack resolver in travis.yml

ryantrinkle:
@tathougies The tests don't seem to succeed on any version of GHC, but I confirmed that it gets as far as 3c0646e6d227bd67cce1d50070515c6e6629c337 for everything (builds on all LTS except 12.0, where it fails with [a deprecation error](https://travis-ci.org/tathougies/beam/jobs/402886100))

tathougies:
I've reworked the entire CI set in beam-0.8.0.0, and am working to move it to master now anyway

PR #293, opened 2018-08-10
Add runPg*ReturningList to support complex Postgres RETURNING queries

ryantrinkle:


tathougies:
I have to get the builds working still, but this looks good otherwise.

tathougies:
@ryantrinkle  would you rebase off current master so CI builds?

PR #310, opened 2018-09-11
Examine all current schemas when generating DB constraints

rimmington:
This fixes the following bug: when running migrations on a schema other than "public" (eg. when running as a non-admin user),  `bringUpToDate` will create the `beam_version` table on first run, but on subsequent runs it will fail to find that table, try to create it again, and fail to do that because the table already exists.


rimmington:
CI failure appears to be unrelated.

tathougies:
Agreed

PR #319, opened 2018-10-14
Add SQLite group_concat function

felixscheinost:
Add the SQLite [group_concat](https://www.sqlite.org/lang_aggfunc.html#groupconcat) function.

I pretty much just copied the [implemention](https://github.com/tathougies/beam/blob/d6b45713de886e7cfbd019eafb86496cc99182dc/beam-postgres/Database/Beam/Postgres/PgSpecific.hs#L1061) of the PG `string_agg` function.

Differences to `pgStringAgg` 

1. Input constraint changed from `IsSqlExpressionSyntaxStringType` to `HasSqlValueSyntax` 

    The reason for that is simply that I need to join an `INTEGER` column in my project. I think a better solution might be to use `cast_`. I couldn't find an implementation of `cast_` though, just `caseE`. Is this a dealbreaker? I guess that e.g. joining `BLOB` really doesn't make sense. Could I constrain the function to take string and number types somehow?

2. Split into two functions beacuse SQLite won't allow `group_concat(distinct col, 'delim')`. Either set quantification or change delimiter.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -12,3 +12,4 @@ Beam would like to thank the following people for their contributions to beam.
 "Brian Jones"<brian@uncannyworks.com>
 "Sam Protas"<sam.protas@gmail.com>
 "Alejandro Duran Pallares"<vwwv@correo.ugr.es>
+"Felix Scheinost" <felix.scheinost@posteo.de>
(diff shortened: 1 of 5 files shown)


tathougies:
Thank you! Please add yourself to the `CONTRIBUTORS` file, and I'll accept this PR :)

Issue #299, opened 2018-08-24
FromField error message is swallowed by Beam


Unisay:
Actual behaviour:

Given the instance:
```haskell
instance FromField MyType where
  fromField _ _ = returnError ConversionFailed f "Could not read value for 'MyType'"
```

The only error message I receive from beam looks like:
```
PgRowParseError (PgRowCouldNotParseField 2)
```
Desired behaviour:

Beam error contains original error message: "Could not read value for 'MyType'"

`beam-core: 0.7.2.2`
`beam-postgres: 0.3.2.2`

runeksvendsen:
I second this.

I'm having a hard time debugging this error, as I can't tell which value the parsing fails for. If `beam` included the `FromField` error message in `PgRowCouldNotParseField`, I could include the `fromField` input value in the `returnError` error message to aid in debugging.

kmicklas:
This is fixed by the change to `BeamRowReadError`.

PR #323, opened 2018-10-21
PeekField: Enable propagating backend errors

runeksvendsen:
Instead of PeekField containing a function that is passed
`Nothing` in case of an error, change the function's type to
`Either [SomeException] a -> f` to allow passing a description
of the error(s) to the backend library.

This allows backends to throw a more descriptive exception
e.g. in case of errors during field parsing.

This PR addresses issue #299


runeksvendsen:
Superseded by #327

PR #327, opened 2018-11-08
Postgres: provide more info when field parsing fails.

zarazek:


zarazek:
I see that there is another pull request providing the same functionality: https://github.com/tathougies/beam/pull/323 Mine is different that it touches only Postgres backend without changing backend interface, so it doesn't break independent backends.

runeksvendsen:
Given that my PR (#323) only affects Postgres anyway, this PR should be preferred, as this PR doesn't alter non-Postgres code. I will close my PR.

kmicklas:
Actually, I forgot this was already fixed by the change to `BeamRowReadError`. I believe the information in `ColumnParseError` should be sufficient now for most debugging.

PR #329, opened 2018-11-26
Improve integer size translations

ryantrinkle:
This branch includes three commits:
* Support `SqlSerial Int16`, `SqlSerial Int32`, and `SqlSerial Int64`
* Change the representation of Int16 to be SMALLINT
* Change the representation of Int to be BIGINT when it is 64 bits wide (but keep it as INTEGER when it is 32 bits wide)

I'm not entirely sure whether the latter two are actually a good idea - personally, I wouldn't use `Int` at all for SQL.  However, it seemed more correct to do it this way.  Let me know if you'd prefer I send these PRs separately or dice it up in some other way.

### DIFF
--- a/beam-migrate/Database/Beam/Migrate/Generics/Tables.hs
+++ b/beam-migrate/Database/Beam/Migrate/Generics/Tables.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE UndecidableInstances #-}
 
@@ -11,6 +12,8 @@ module Database.Beam.Migrate.Generics.Tables
 
   , HasNullableConstraint, NullableStatus
   ) where
+
+#include "MachDeps.h"
 
 import Database.Beam
 import Database.Beam.Backend.SQL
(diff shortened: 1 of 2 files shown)


tathougies:
Would you be able to rebase on top of beam-0.8.0.0 (recently merged beam-0.8.0.0-migrate into that) and then push again. That should get CI going... sorry for the delay!

kmicklas:
Basing the representation of `Int` on the machine word size seems like it could be a recipe for nasty surprises. I'd be in favor of leaving it as 32 bits always, since that guarantees deserializing from the database will always work, and if you're actually trying to store large integers you would hopefully know to use `Int64`.

ryantrinkle:
@kmicklas Why even keep Int around if it's going to be wrong?

kmicklas:
@ryantrinkle Yeah I'd prefer removing the `Int` instances entirely, but it's likely to break enough code that I'll defer to @tathougies on that. We could maybe alleviate the pain a bit with some custom type errors saying "please use `Int32`". Either way, I think that should go in a separate PR from the first two commits.

3noch:
> Yeah I'd prefer removing the Int instances entirely, but it's likely to break enough code that I'll defer to @tathougies on that. 
People can just add it back as an orphan to avoid breakage if they really want to.

kmicklas:
Merging this as is. I'll consider removing the `Int` instances before the next release.

PR #330, opened 2018-11-27
Update the no-primary-key example

adetokunbo:
To reflect the tested code added in 8503c45e118d911b5d158d23f9b25fcc0f3acb37


tathougies:
Thank you!

PR #335, opened 2018-12-10
Add indices support (WIP)

Martoon-00:
## Motivation

Currently, there is no way to deal with the database schema purely via haskell code, which is not good at least because when we define our schema in raw SQL, we have table and fields names defined in several places.
I'm planning to make SQL engine know about foreign key constraints in the next PR, and these two seem to provide a basis for moving away from manual schemas definition completely.

## PR Details

The following features are implemented:
- [x] Manual indices definition.
Here a user can provide on the per-table basis a list of indices, each index is defined by a sequence of its fields and its settings (only most basic settings are supported for now). With syntax as the [following](https://github.com/serokell/beam/blob/martoon%2Fadd-indices/beam-core/test/Database/Beam/Test/Schema.hs#L440).

- [x] Automatic indices detection.
A user can make a `DatabaseIndices` (which is similar to `DatabaseSettings`, but contains information about indices correspondingly) via `defaultDbIndices`, which gathers indices on those columns which are assumed to participate in `JOIN`s.
Later the user can add these indices to the custom ones via dedicated function `mergeDbIndices`, or immediattely use `addDefaultDbIndices`.

- [x] Manual/automatic migrations support.
Previously gathered indices are converted to `TableChecks` and added to `CheckedDatabaseSettings`.
Two action providers are supplied: `ADD INDEX` and `DROP INDEX`.

Herewith, unfortunately, there is no sqlite support, since there is no way to modify indices of existing table there. Common workaround is to rename the table, create a new one with desired indices and copy all the data, but I'm not yet ready to write copying action providers, although generally that scheme seems to be perfectly discoverable with existing automatic migration means.

- [ ] Testing.
  - [x] Unit tests on indices definition.
  - [ ] Manual migrations.
  - [ ] Automatic migrations.

- [ ] Haddock: builds.

- [ ] Manual: should I extend it?

### Notes

I have a set of `TODO`s remained whether I'm not sure about proper design choice, would appreciate hearing your point of view on this.

Also I added `.stylish-haskell` which corresponds to the style of this project, otherwise I get pretty mad diffs each time I save a file.


tathougies:
Hello!

Wow.. Thanks! This is a lot to digest, but right off the bat, I'll ask you to rebase on top of the beam-0.8.0.0 branch, since that represents the next release. There'll be a few changes you'll have to make to get it compiling. The upside is that it should trigger a gitlab CI build, which will have to pass before I merge.

Additions to the manual would be nice.

I'm still getting through the code.

Travis

beam-core/Database/Beam/Schema/Indices.hs
```diff
@@ -0,0 +1,357 @@
...
+import           Data.Text (Text)
+import qualified Data.Text as T
+
+import           GHC.Exts (Constraint, fromList)
+import           GHC.Generics hiding (C, R)
+import           GHC.TypeLits
+
+import           Database.Beam.Schema.Tables
+
+-- Some day it should have more options and allow to modify them depending on the backend.
```
tathougies: When merged, add a ticket for this to the issue reporting system

beam-core/Database/Beam/Schema/Indices.hs
```diff
@@ -0,0 +1,357 @@
...
+
+-- | Traverses fields of the given table and builds indices for all encountered 'PrimaryKey's.
+class AutoEntityIndex be db tbl where
+    autoEntityIndices :: EntityIndices be db tbl
+
+-- Other types of entities are approaching, and we probably don't want to define
+-- instances for all of them.
+-- | Traverses the given table and for every field which is some 'PrimaryKey'
+-- makes corresponding SQL index, this allows "JOIN"s on this table perform nicely.
+instance {-# OVERLAPPABLE #-}
```
tathougies: Why are these instances overlappable?

beam-core/Database/Beam/Schema/Indices.hs
```diff
@@ -0,0 +1,357 @@
...
+
+-- * Automatic indices definition
+
+-- | Indicates a reference from table @tbl@ to table @tbl'@.
+data (tbl :: (* -> *) -> *) :-> (tbl' :: (* -> *) -> *)
+
+-- | Provide options for an automatically created index, which is caused by a primary key
+-- of one table being embedded into another.
+-- This typeclass is only needed to be defined for inter-table references (see ':->').
+class IndexFromReference reference where
```
tathougies: The issue I see right now is that tables should be able to have two foreign keys that refer to the same table type at the haskell level, but really refer to two different tables at the database level.

For example

```haskell
data TwoKeyT f =
  TwoKey {
    _key1 :: PrimaryKey RefT f
    _key2 :: PrimaryKey RefT f
  }

data RefT f = Ref { _field1 :: C f Text, _field2 :: C f Int }

data Database entity = Database { _ref1 :: entity (TableEntity RefT), _ref2 :: entity (TableEntity RefT), _twoKey :: entity (TableEntity TwoKeyT) }
```

How would I make it so that it knows that _key1 refers to _ref2 and _key2 to _ref1 (or vice versa). I believe the way it is now, each type is uniquely linked to one database table, which is a large constraint.

One way to get around this that would also (I think) have the benefit of removing the instance overlap is to not use the type class mechanism and require the user to construct an appropriate value.

Martoon-00:
Oh, having CI checking my work would be really nice!
I see `beam-0.8.0.0-fix-ci` branch and multiple `travis/beam-0.8.0.0-*` braches, which one should I choose?

tathougies:
Just regular beam-0.8.0.0. I'll get rid of the other ones. They're all merged.

beam-core/Database/Beam/Schema/Indices.hs
Martoon-00: Well, I'm not sure whether is it a good choice, but I saw that there are plans to support many different entities type, while we want to have indices only for `TableEntities`. It looked like the one who will enable support for those other entities should not bother about indices at all, thus a default overlappable instance.
Should I define instances for each entity type explicitly instead?

beam-core/Database/Beam/Schema/Indices.hs
Martoon-00: I agree this is a major problem. In case of this implementation via type class mechanism a workaround is to parametrize table type with a phantom type parameter, thus making different tables of the same type distinguishable in a Haskell code.
However, making user explicitly make a construct which describes options for the generated indices seems to be virtually better in several ways. Doing so may be tricky if we still want the type system to enforce the user to specify indices for all embedded `PrimaryKey`s present; I described my raw suggestion right below this code, we could discuss it :)

Martoon-00:
That's weird, I see no plain `beam-0.8.0.0` branch :thinking:

tathougies:
This one: https://github.com/tathougies/beam/tree/travis/beam-0.8.0.0

tathougies:
The latest commit there should be 377596e

Martoon-00:
Oh right, I searched in my forked repository. Thanks for helping with figuring this out, I'm rebasing.

tathougies:
You'll need to resolve conflicts too to get CI working.

Martoon-00:
Yeah, it's gonna take a while :)

Martoon-00:
Finally, I rebased upon beam-0.8.0.0 branch. Now I get a couple of small issues:
1. CI is not triggered, probably because I'm proposing this change from another repo.
Do you know how to / can enable that?
2. I activated CI in my fork for now, and sometimes tests fail https://travis-ci.org/serokell/beam/jobs/470453508.
Postgres seems to happily truncate `\0` character in `text`s.

tp-woven:
Oh, wow, this looks really cool. Any chance of this getting resurrected and merged?

kmicklas:
@tp-woven You're welcome to take it over! I'm happy to review but unfortunately don't have the time for major feature development on Beam these days.

Martoon-00:
Yeah, @tp-woven I would be glad if you manage to take this work or pick it as inspiration for your own implementation, especially since @kmicklas approved this. Now I cannot allocate time to this.

If I remember correctly, this PR mostly handles the core part (not tested at all) but may need a decent update of the migration package.

tp-woven:
Thanks, I'll try to take over. Fair warning: I'm a Haskell noob, so it may take a while... 😅

PR #339, opened 2018-12-24
Update to lts-13.0

JorisM:
these are the changes that I had to made to get it to work with the new lts.


tathougies:
Thank you!

PR #347, opened 2019-01-24
Add HasInsertOnConflict class for postgres/sqlite-like upsert support

kmicklas:
I'd like to add upsert support for `beam-sqlite` and this seemed like a good opportunity to abstract across backends. Before I finish that I figured I'd get this out to hear your thoughts on the design.

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
+++ b/beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
@@ -11,7 +11,7 @@ module Database.Beam.Backend.SQL.BeamExtensions
   ( MonadBeamInsertReturning(..)
   , MonadBeamUpdateReturning(..)
   , MonadBeamDeleteReturning(..)
-  , HasInsertOnConflict(..)
+  , BeamHasInsertOnConflict(..)
 
   , SqlSerial(..)
   , onConflictUpdateInstead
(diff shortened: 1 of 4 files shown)


tathougies:
Can this be implemented on SQLite?

kmicklas:
I haven't tried it, but it exists: https://www.sqlite.org/lang_UPSERT.html

3noch:
> UPSERT in SQLite follows the syntax established by PostgreSQL.

Well that's convenient.

tathougies:
I am a fan. The only thing is I would add the word 'beam' somewhere into the name of the typeclass `HasInsertOnConflict`, so it is clear that this is a beam extension.

kmicklas:
@tathougies The other issue I know of is that the instance currently lives in `Database.Beam.Postgres.Full`, which is not automatically imported as part of `Database.Beam.Postgres`, so it's a true orphan. What module do you think would be most appropriate for it to live in?

tathougies:
I think you should be able to move it to `Database.Beam.Postgres`.

kmicklas:
@tathougies I've renamed the class and moved the instance to `Database.Beam.Postgres`.

kmicklas:
@tathougies Bump.

kmicklas:
Hey @tathougies are there any issues with this pull request? I'd like to start implementing the SQLite support soon but I wanted to make sure you were okay with this interface first.

tathougies:
No problems with this. Let me look at it one last time and then I'll merge it. Thanks for the contribution!

PR #357, opened 2019-03-13
export modifyEntityName, modifyTableFields

templateK:
ghc complains `modifyTable` is deprecated. but `Database.Beam` doesn't  export alternative functions.

```
<interactive>:1:1: warning: [-Wdeprecations]
    In the use of ‘modifyTable’
    (imported from Database.Beam, but defined in Database.Beam.Schema.Tables):
    Deprecated: "Instead of 'modifyTable fTblNm fFields', use 'modifyEntityName _ <> modifyTableFields _'"
```


tathougies:
Should actually be fixed in latest master.

templateK:
thanks for quick response. 👍

PR #365, opened 2019-03-31
Use the Database category in hackage

mightybyte:
This make it so this package shows up next to all the other beam packages.


tathougies:
thank you!

PR #374, opened 2019-05-11
WIP: Allow databases to be nested

ryantrinkle:
I think it would be good to rename the inner tables, but this hasn't been done yet


tathougies:
I think this looks good. It would be nice to have a test in the schema tests portion of `beam-core`, if possible.

kmicklas:
Closing in favor of https://github.com/tathougies/beam/pull/423.

PR #378, opened 2019-05-16
Implement upsert support for SQLite

kmicklas:


beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
```diff
@@ -146,6 +146,10 @@ class BeamSqlBackend be => BeamHasInsertOnConflict be where
   type SqlConflictTarget be (table :: (* -> *) -> *) :: *
   type SqlConflictAction be (table :: (* -> *) -> *) :: *
 
+  -- | This is necessary for places that we can't use a forall because the
```
tathougies: You can probably use `QInternal` here?

kmicklas: Done. Out of curiosity, what's the theory behind `QInternal`? It doesn't look used anywhere. Should all instances of `PostgresInaccessible` be replaced with that?

tathougies: It's only accessible in beam-core, I think?

PR #390, opened 2019-07-27
Add access to excluded row to onConflictUpdateSetWhere

kmicklas:


kmicklas:
@tathougies Note that this conflicts with https://github.com/tathougies/beam/pull/388 and the conflict will be easier to resolve if that PR is merged first (since this one is so small).

PR #391, opened 2019-08-01
Adds unique checks to `getDbConstraints` in Postgres

ProofOfKeags:


CaptJakk:
This is actually super naive, it will fail for any table names that are multiword, looking into a better way to ascertain column/table names besides text splitting.

DO NOT MERGE

CaptJakk:
This should be good now. It more intelligently deals with the issue of deducing table names from the uniqueness constraints. I tested this with the use case that motivated me to add this in the first place, but am slightly unsure how we should go about testing this in general. Open to suggestions.

dfordivam:
This code doesn't work for me, though this does (ie the value we get from query unmodified)
```
         pure tblUniqs'
```

dfordivam:
query output for my case 
```
SELECT cl.relname, conname FROM pg_constraint co JOIN pg_namespace na ON co.connamespace = na.oid JOIN pg_class cl on cl.relfilenode= co.conrelid WHERE co.contype = 'u';
   relname   |    conname
-------------+---------------
 appDb_users | user_username
```

CaptJakk:
Hmm. What version of Postgres?

CaptJakk:
so the main curiosity I have here is why your relname is prefixed with "appDb_" and why all of my uniqueness constraints had "_key" suffixed onto them. I can't seem to find a query that can definitively just get the table and column for uniqueness constraints. It seems that uniqueness constraints can be named whatever you want them to be and there is no obvious way for me to get the actual column name and table name from that query. Would you be willing to share the model definition that you have that caused this to happen?

CaptJakk:
Looks like the following query might yield some valuable progress here:
```
select table_name, column_name from information_schema.constraint_column_usage ccu where ccu.constraint_name in (SELECT co.conname FROM pg_constraint co JOIN pg_namespace na ON co.connamespace = na.oid JOIN pg_class cl on cl.relfilenode= co.conrelid WHERE co.contype = 'u');
```

Please let me know if that works for you, If it does I will update the PR to use that.

dfordivam:
> so the main curiosity I have here is why your relname is prefixed with "appDb_" and why all of my uniqueness constraints had "_key" suffixed onto them. I can't seem to find a query that can definitively just get the table and column for uniqueness constraints. It seems that uniqueness constraints can be named whatever you want them to be and there is no obvious way for me to get the actual column name and table name from that query. Would you be willing to share the model definition that you have that caused this to happen?

I added the unique constraint manually, as the `autoMigrate` doesn't seem to support that at the moment.
```
data UserT f = User
  { _user_id       :: !(C f (SqlSerial Int))
  , _user_password :: !(C f Text)
  , _user_email    :: !(C f Text)
  , _user_username :: !(C f Text)
  }
data AppDb f = AppDb
  { _appDb_users       :: f (TableEntity UserT)
  } deriving (Generic)
ALTER TABLE "appDb_users" ADD CONSTRAINT user_username UNIQUE ("user_username") ;
```

CaptJakk:
Did you try the query I put in my last comment? I think it should be more robust.

dfordivam:
> Did you try the query I put in my last comment? I think it should be more robust.

This gives exactly the same output as before. My usage of Unique might be a little different to the standard, as I added the constraint myself. But it also points to a potential problem in the code.

CaptJakk:
It looks like the output is consistent with your model though. What would
be the correct output in your case?

CaptJakk:
I'd like to revisit this. Did you check the query I left in the comments to see if it fixed the issues associated with the original PR?

dfordivam:
@CaptJakk did you ask me?
As I said the select query you gave, and the one in PR, both gives same output. So they both wont work for my case. 
In my case I did not need any processing of the query output.

ProofOfKeags:
Closing this as it doesn't seem relevant to anyone at this point.

PR #394, opened 2019-08-16
Clarify how to rename primary key fields.

ZzAntares:
IMHO Although the types lead you to be able to rename primary key fields its better to be explicit about it.

### DIFF
--- a/docs/tutorials/tutorial2.md
+++ b/docs/tutorials/tutorial2.md
@@ -165,16 +165,19 @@ value.
 
     `_addressLine1 = "address1"`
 
-When renaming the name of a field referring to a foreign key (for example
-the `_addressForUser` field), remember to specify the field name as a
-`PrimaryKey` field like:
+When renaming a field referring to a foreign key (for example the
+`_addressForUser` field), remember to wrap the field name with the table's
+`PrimaryKey` constructor:
 
 `_addressForUser = UserId "user"`
 
-Given that in part one we defined `UserId` as a type alias:
+Given that in part one we defined `UserId` as the `PrimaryKey` constructor for
+`UserT`:
 
 ```haskell
-type UserId = PrimaryKey UserT Identity
+instance Table UserT where
+   data PrimaryKey UserT f = UserId (Columnar f Text) deriving (Generic, Beamable)
+   primaryKey = UserId . _userEmail
 ```
 
 If you didn't need to modify any of the field names, you can omit
--- a/docs/tutorials/tutorial2.md
+++ b/docs/tutorials/tutorial2.md
@@ -165,7 +165,7 @@ value.
 
     `_addressLine1 = "address1"`
 
-When renaming the name of a field referring to a primary key (for example
+When renaming the name of a field referring to a foreign key (for example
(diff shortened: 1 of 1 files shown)


docs/tutorials/tutorial2.md
```diff
@@ -165,6 +165,18 @@ value.
 
     `_addressLine1 = "address1"`
 
+When renaming the name of a field refering to a primary key (for example
```
3noch: referring

docs/tutorials/tutorial2.md
```diff
@@ -165,6 +165,18 @@ value.
 
     `_addressLine1 = "address1"`
 
+When renaming the name of a field referring to a primary key (for example
```
kmicklas: I'd say "referring to a foreign key" (even though you use `PrimaryKey` for this) as primary key suggests the key of the table in question.

docs/tutorials/tutorial2.md
```diff
@@ -165,6 +165,18 @@ value.
...
+the `_addressForUser` field), remember to specify the field name as a
+`PrimaryKey` field like:
+
+`_addressForUser = UserId "user"`
+
+Given that in part one we defined `UserId` as a type alias:
+
+```haskell
+type UserId = PrimaryKey UserT Identity
```
kmicklas: I think you're confusing the type synonym name and the constructor name here. They're not inherently related and this type synonym isn't what's used in the table definition (as it's not parametric over the `f` parameter).

ZzAntares: Hi, I see what you mean, but I was just quoting the definition as written in part one of the tutorial, assuming the reader is following the guide that is the type synonym you have up to that point. Are you suggesting we change the definition also in part one? something like `type UserPk f = PrimaryKey UserT f`?

kmicklas: The relevant definition is the constructor name here:

```haskell
instance Table UserT where
   data PrimaryKey UserT f = UserId (Columnar f Text) deriving (Generic, Beamable)
   primaryKey = UserId . _userEmail
```

The fact that there is also a type synonym for `PrimaryKey UserT Identity` and that it shares the name `UserId` isn't why you need to use `UserId` to specify the name of the foreign key column.

docs/tutorials/tutorial2.md
```diff
@@ -165,6 +165,18 @@ value.
 
     `_addressLine1 = "address1"`
 
+When renaming the name of a field referring to a foreign key (for example
+the `_addressForUser` field), remember to specify the field name as a
+`PrimaryKey` field like:
+
+`_addressForUser = UserId "user"`
```
kmicklas: I think this will render better as either a code block or inline with the previous paragraph. (I have a slight preference for the latter.)

docs/tutorials/tutorial2.md
ZzAntares: oooh I get it know

fosskers:
Following the suggestions here, I was able to change my code and make some FK renamings finally compile. Thank you!

PR #405, opened 2019-09-28
Ghc 8.8

phadej:


beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
```diff
@@ -181,7 +182,7 @@ class BeamSqlBackend be => BeamHasInsertOnConflict be where
 
 newtype InaccessibleQAssignment be = InaccessibleQAssignment
   { unInaccessibleQAssignment :: [(BeamSqlBackendFieldNameSyntax be, BeamSqlBackendExpressionSyntax be)]
-  } deriving (Semigroup, Monoid)
+  } deriving (Data.Semigroup.Semigroup, Monoid)
```
kmicklas: Why is this fully qualified?

beam-core/Database/Beam/Backend/SQL/Row.hs
```diff
@@ -25,6 +25,8 @@ import           Data.Typeable
 import           Data.Vector.Sized (Vector)
 import qualified Data.Vector.Sized as Vector
 
+import qualified Control.Monad.Fail as Fail
```
kmicklas: Why is this imported qualified?

beam-core/beam-core.cabal
```diff
@@ -56,29 +56,32 @@ library
                        Database.Beam.Query.Relationships
 
                        Database.Beam.Schema.Lenses
-  build-depends:       base         >=4.7     && <5.0,
+  build-depends:       base         >=4.9     && <5.0,
```
kmicklas: Are the old lower bounds wrong or something? Ideally we wouldn't have to raise them just to support a barely released GHC.

beam-migrate/Database/Beam/Migrate/Log.hs
```diff
@@ -143,7 +145,7 @@ recordCommit commitId = do
 
 -- Ensure the backend tables exist
 ensureBackendTables :: forall be m
-                     . BeamSqlBackendCanSerialize be Text
+                     . (BeamSqlBackendCanSerialize be Text, Fail.MonadFail m)
```
kmicklas: Won't this need some CPP to not break with GHC 8.4 and earlier?

stack.yaml
```diff
@@ -13,3 +13,7 @@ packages:
 
 nix:
   shell-file: shell.nix
+
+#extra-deps:
```
kmicklas: What are these for?

beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
phadej: To suppress unused import warning on GHC-8.8

beam-core/Database/Beam/Backend/SQL/Row.hs
phadej: Because on GHC-8.6 (and older) `fail` and `Fail.fail` are different

beam-core/beam-core.cabal
phadej: `beam-core` requires GHC-8.0, which comes with `base-4.9`.

Other bounds are unsatisfiable as well.

beam-migrate/Database/Beam/Migrate/Log.hs
phadej: The three packages I mentioned compile (and test pass, except the ones requiring PostgreSQL server) with GHC-8.0 ... GHC-8.8

This is breaking interface change, but there are no way around it

stack.yaml
phadej: If you try to compile with updated (but commented out) `nightly-2019-09-28`, you'll need to add these.

beam-core/beam-core.cabal
tathougies: Does it not compile with the lower bound? That should be a bug filed against cabal or stack.

phadej: @tathougies I don't understand your comment? `>=4.7` is "wrong" as it gives impression that `beam-core` should be buildable with GHC-7.8 (base-4.7), which isn't the case.

phadej:
Is there anything I can do to advance this PR forward?

beam-core/beam-core.cabal
tathougies: So beam no longer would compile on 7.8, due to the changes in base, is that right? That's fine with me of course

phadej: You started using `TypeApplications` which is GHC-8.0+ language extension.

tathougies: Good point!

PR #417, opened 2020-01-08
SomeBeamMigrationBackend needs MonadFail

Ericson2314:
Otherwise the migration, which can fail, cannot be run.

### DIFF
--- a/beam-migrate/Database/Beam/Migrate/Backend.hs
+++ b/beam-migrate/Database/Beam/Migrate/Backend.hs
@@ -74,6 +74,7 @@ type DdlError = String
 data BeamMigrationBackend be m where
   BeamMigrationBackend ::
     ( MonadBeam be m
+    , Fail.MonadFail m -- TODO check whether this is good
     , HasQBuilder be
     , BeamMigrateSqlBackend be
     , HasDataTypeCreatedCheck (BeamMigrateSqlBackendDataTypeSyntax be)
@@ -96,10 +97,7 @@ data BeamMigrationBackend be m where
 -- | Monomorphic wrapper for use with plugin loaders that cannot handle
 -- polymorphism
 data SomeBeamMigrationBackend where
-  SomeBeamMigrationBackend :: ( BeamMigrateSqlBackend be
-                              , Typeable be
-                              , Fail.MonadFail m -- TODO check whether this is good
-                              )
+  SomeBeamMigrationBackend :: Typeable be
                            => BeamMigrationBackend be m
                            -> SomeBeamMigrationBackend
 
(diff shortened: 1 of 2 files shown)


beam-migrate/Database/Beam/Migrate/Backend.hs
```diff
@@ -96,7 +97,9 @@ data BeamMigrationBackend be m where
 -- polymorphism
 data SomeBeamMigrationBackend where
   SomeBeamMigrationBackend :: ( BeamMigrateSqlBackend be
-                              , Typeable be )
+                              , Typeable be
+                              , Fail.MonadFail m -- TODO check whether this is good
```
kmicklas: Is this needed? It seems odd when there wasn't even a `Monad` constraint before.

Ericson2314: Hmm now that you mention it I think `MonadFail` could be moved to `BeamMigrationBackend` (scroll up) and `BeamMigrateSqlBackend` can be deleted as its already included in that list. I'm a bit skeptical of all these dictionaries (explicit or implicit), so maybe there is more that could be done but it would a much more invasive.

Taking a step back, the underlying problem is migrations can fail, and `ensureBackendTables`, which is used by default in particular, does call `fail` after the 8.8 fixes. We can use something more structured than `Fail` (always fine by me!) but we need something. I don't know how failure usually works with beam, so I can't answer much better.

Ericson2314: Moved the constraints, testing now

Ericson2314: Testing in https://github.com/obsidiansystems/beam/pull/4

Ericson2314: OK it passed.

beam-migrate/Database/Beam/Migrate/Backend.hs
```diff
@@ -95,8 +97,7 @@ data BeamMigrationBackend be m where
 -- | Monomorphic wrapper for use with plugin loaders that cannot handle
 -- polymorphism
 data SomeBeamMigrationBackend where
-  SomeBeamMigrationBackend :: ( BeamMigrateSqlBackend be
-                              , Typeable be )
+  SomeBeamMigrationBackend :: Typeable be
```
kmicklas: Can you put the `BeamMigrateSqlBackend` constraint back?

beam-migrate/Database/Beam/Migrate/Backend.hs
```diff
@@ -73,6 +74,7 @@ type DdlError = String
 data BeamMigrationBackend be m where
   BeamMigrationBackend ::
     ( MonadBeam be m
+    , Fail.MonadFail m -- TODO check whether this is good
```
kmicklas: Can you remove the TODO? I agree that it would be nice to have something more structured but I think this is correct for now.

beam-migrate/Database/Beam/Migrate/Backend.hs
Ericson2314: It's not needed because you get it when you unpack the thing inside.

PR #421, opened 2020-01-13
Enhance Database.Beam.Migrate.Simple

lpsmith:
From https://github.com/obsidiansystems/beam/commit/5f7396505488ba7ba250993dbf1b8ea8f86c1805


tathougies:
Hello there... sorry for the long delay. I think this is a good change, but I worry that `autoMigrate` will now not fail on an unsafe migration.

kmicklas (commented):
@tathougies Unless I'm missing something, it looks like `autoMigrate` will still fail on an unsafe migration.

PR #422, opened 2020-01-13
Add instance HasDefaultDatatype Postgres UTCTime

lpsmith:
From https://github.com/obsidiansystems/beam/commit/708ad2f8b1f1b5991a019c7a3c1e7ce1dc92f642

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Types.hs
+++ b/beam-postgres/Database/Beam/Postgres/Types.hs
@@ -170,6 +170,9 @@ instance HasDefaultSqlDataType Postgres ByteString where
 instance HasDefaultSqlDataType Postgres LocalTime where
   defaultSqlDataType _ _ _ = timestampType Nothing False
 
+instance HasDefaultSqlDataType Postgres UTCTime where
+  defaultSqlDataType _ _ _ = timestampType Nothing True
+
 instance HasDefaultSqlDataType Postgres (SqlSerial Int) where
   defaultSqlDataType _ _ False = pgSerialType
   defaultSqlDataType _ _ _ = intType


3noch:
c.f. https://github.com/tathougies/beam/pull/283

3noch:
Does this end up using something akin to `time AT TIME ZONE 'UTC'`?

lpsmith:
What do you mean?  This specifies, `timestamp with time zone`,  which is definitely the best match for `UTCTime` currently in postgres.    `postgresql-simple` specifies `Z` in the data format itself, so yes, the interpolated value is always interpreted as UTC time.   (And hasql goes the binary route,  which means you are obligated to send UTC time for `timestamp with time zone`)

The biggest issue with that matchup is the fact that postgres claims it's built-in `timestamp with time zone` is UTC time (which has leap seconds) but the data format doesn't track the date seperately from time of day,  which any fully correct representation of UTC time must do. (Haskell's time library gets one up on postgresql here,  but loses out in other ways too.) 

So postgres's built in time types basically pretend that leap seconds don't exist, like most software (sadly) does.  I'd like to fix this someday, offering true `taitime` and `utctime` types.   But I can't say it's a high priority for me.

3noch:
I meant that when selecting this column, the SQL effectively does `x AT TIME ZONE 'UTC'` which is only valid on columns that have time zones.

lpsmith:
Yes, `timestamp with time zone` is always returned with sufficient information that,  if it's not already in UTC time,  it's possible to convert it to UTC time,  so yes, postgresql-simple has you covered on that count.  (And the binary format is _always_ utc,  and you have to convert it to a localtime yourself.)

lpsmith:
Also, `timestamp with time zone` stores neither offsets from UTC nor timezone information proper.

3noch:
Awesome.

fosskers:
This would be nice to have.

fosskers:
Any reason this can't go in?

kmicklas:
@tathougies do your objections to this instance on https://github.com/tathougies/beam/pull/283 apply here?

fosskers:
I pasted the instance given here as an orphan in my own code, and everything seems to be working as expected.

PR #423, opened 2020-01-13
Allow databases to be nested

lpsmith:
From https://github.com/obsidiansystems/beam/commit/1d58f21485e5b95e1edb662d3de26f5d3572289d

### DIFF
--- a/beam-migrate/Database/Beam/Migrate/Generics/Types.hs
+++ b/beam-migrate/Database/Beam/Migrate/Generics/Types.hs
@@ -2,7 +2,9 @@
 module Database.Beam.Migrate.Generics.Types where
 
 import           Database.Beam.Migrate.Types
+import           Database.Beam.Schema.Tables
 
+import           Data.Functor.Identity
 import           Data.Proxy
 import qualified Data.Text as T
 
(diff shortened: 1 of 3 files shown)


beam-migrate/Database/Beam/Migrate/Generics/Types.hs
```diff
@@ -27,3 +29,33 @@ instance ( Selector f, IsCheckedDatabaseEntity be x
...
     where name = T.pack (selName (undefined :: S1 f (Rec0 (CheckedDatabaseEntity be db x)) ()))
+
+instance ( Selector f
+         , Generic (CheckedDatabaseSettings be innerDb)
+         , GAutoMigratableDb be (Rep (CheckedDatabaseSettings be innerDb))
+         , Database be innerDb
+         ) =>
+  GAutoMigratableDb be (S1 f (Rec0 (innerDb (CheckedDatabaseEntity be outerDb)))) where
+
+  defaultMigratableDbSettings' _ = M1 (K1 (runIdentity $ zipTables (Proxy :: Proxy be) (\_ -> pure . changeCheckedDatabaseEntityDb @innerDb @outerDb @be) settings settings :: innerDb (CheckedDatabaseEntity be outerDb))) --TODO: Rename inner tables based on field name
```
kmicklas: At the very least this `TODO` needs to be removed. I'm not super convinced one way or another that inner tables should be prefixed with the container field name, but whatever way we go with we should stick with as users depend on it for migrations to work.

ryantrinkle: This has been resolved

ryantrinkle:
@kmicklas @tathougies  It's been a while, but I believe this may be mergeable now!

One thing I did not do here: ideally entities would use `[Text]` rather than `Text` for their names internally just like fields do.  However, this would be a breaking change if anyone uses this functionality.  Let me know if you'd prefer I add that here.

ryantrinkle:
Sorry guys, I do see that you asked for tests.  I'll add that and mark this ready when that's done.

kmicklas:
Thanks @ryantrinkle! I'll be on vacation the next week but I can review when I get back.

Switching to `[Text]` is probably a good idea but I'd prefer that be a separate PR.

kmicklas:
Replaced with #661. Feel free to follow up with any missing features.

PR #433, opened 2020-02-13
Re-export a few missing functions

mightybyte:
It looks like these functions got moved to a different module where everything was exported, then moved back but not added to the the exports for this module.

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Full.hs
+++ b/beam-postgres/Database/Beam/Postgres/Full.hs
@@ -37,6 +37,8 @@ module Database.Beam.Postgres.Full
   , conflictingConstraint
   , onConflictUpdateSet
   , onConflictUpdateSetWhere, onConflictUpdateInstead
+  , conflictingFields, conflictingFieldsWhere
+  , onConflictDoNothing
 
   -- * @UPDATE RETURNING@
   , PgUpdateReturning(..)
--- a/beam-postgres/Database/Beam/Postgres/Full.hs
+++ b/beam-postgres/Database/Beam/Postgres/Full.hs
@@ -35,8 +35,6 @@ module Database.Beam.Postgres.Full
 
   , onConflictDefault, onConflict
   , conflictingConstraint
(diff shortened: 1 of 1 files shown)


kmicklas:
Can you add back `anyConflict` and `onConflictSetAll` as well? My mistake for removing all of these: https://github.com/tathougies/beam/pull/347/files#diff-621691a5fa5afd0286d2fe3e18fd1360

mightybyte:
@kmicklas `anyConflict` looks like it's a part of a type class, so I think nothing needs to be done for that?  And it looks like `onConflictSetAll` does not exist any more.

kmicklas:
@mightybyte `conflictingFields`, `conflictingFieldsWhere`, and `onConflictDoNothing` are part of the same type class as `anyConflict`. `Database.Beam.Postgres.Full` should probably just re-export `BeamHasInsertOnConflict(..)` so all the methods are covered, as well as `onConflictUpdateAll` and `onConflictUpdateInstead` which are outside the class but provide functionality that was in `Database.Beam.Postgres.Full` before #347.

mightybyte:
Updated with those changes

beam-postgres/Database/Beam/Postgres/Full.hs
```diff
@@ -36,7 +36,10 @@ module Database.Beam.Postgres.Full
   , onConflictDefault, onConflict
   , conflictingConstraint
   , onConflictUpdateSet
-  , onConflictUpdateSetWhere, onConflictUpdateInstead
+  , onConflictUpdateSetWhere,
```
kmicklas: `onConflictUpdateSet` and `onConflictUpdateSetWhere` are both already included in `BeamHasInsertOnConflict(..)`.

Sorry to keep going back and forth like this. :smile:

mightybyte: Updated

kmicklas:
Thanks!

Issue #431, opened 2020-01-29
Issues with DOUBLE PRECISION and Migration


fosskers:
Hi there, I seem to have hit [something related to this PR](https://github.com/tathougies/beam/pull/240). I'm running with the following:
```
beam-core 0.8.0.0
beam-migrate 0.4.0.1
beam-sqlite 0.4.0.0
```
I have a table with a `Double` field (among other things). When I auto-create this table via `createSchema` or `autoMigrate` everything goes fine. Any subsequent attempts result in:
```
[(TableHasColumn Sqlite: Table QualifiedName Nothing "transactions" must have a column "transaction_gasPrice" of DOUBLE PRECISION)]
```
Note that a dump of the offending table within `sqlite3` gives:
```
sqlite> .dump transactions
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
CREATE TABLE IF NOT EXISTS "transactions"("transaction_id" INTEGER NOT NULL , "transaction_chainId" INTEGER NOT NULL , "transaction_block__block_id" INTEGER NOT NULL , "transaction_creationTime" INTEGER NOT NULL , "transaction_ttl" INTEGER NOT NULL , "transaction_gasLimit" INTEGER NOT NULL , "transaction_gasPrice" DOUBLE PRECISION NOT NULL , "transaction_sender" VARCHAR NOT NULL , "transaction_nonce" VARCHAR NOT NULL , "transaction_requestKey" VARCHAR NOT NULL , PRIMARY KEY("transaction_id"));
COMMIT;
```
If I remove that `Double` field, the problem goes away and the `autoMigrate` succeeds every time.
Any thoughts? Thank you.

fosskers:
Hi there, any thoughts on this?

kmicklas:
I can confirm that the column gets created with type `DOUBLE PRECISION` but the constraint gets picked up as just `DOUBLE`:

```
TableHasColumn Sqlite: Table QualifiedName Nothing "test" must have a column "test" of DOUBLE
```

fosskers:
For the record I'll state here that this issue happens only with the sqlite backend, and isn't present at all after a switch to postgres.

kmicklas:
Yeah, it looks like an issue with `parseSqliteDataType`. I can try to put up a fix today.

PR #436, opened 2020-02-15
Fix emitting of SQL 92 DECIMAL type

kmicklas:
`DOUBLE` is not a valid SQL 92 data type, and Postgres completely rejects it. SQLite would misleadingly accept it with `REAL` affinity instead of `NUMERIC` which is what you might expect if a precision were specified.

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL/Builder.hs
+++ b/beam-core/Database/Beam/Backend/SQL/Builder.hs
@@ -433,7 +433,7 @@ instance IsSql92DataTypeSyntax SqlSyntaxBuilder where
     varBitType prec = SqlSyntaxBuilder ("BIT VARYING" <> sqlOptPrec prec)
 
     numericType prec = SqlSyntaxBuilder ("NUMERIC" <> sqlOptNumericPrec prec)
-    decimalType prec = SqlSyntaxBuilder ("DOUBLE" <> sqlOptNumericPrec prec)
+    decimalType prec = SqlSyntaxBuilder ("DECIMAL" <> sqlOptNumericPrec prec)
 
     intType = SqlSyntaxBuilder "INT"
     smallIntType = SqlSyntaxBuilder "SMALLINT"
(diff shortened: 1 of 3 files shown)


kmicklas:
@tathougies This seems like just a bug, but does it have implications for people already using `beam-migrate`? I can't imagine it would break anything using Postgres since `decimalType` couldn't have possibly worked emitting `DOUBLE`. Maybe for SQLite it could break existing migrations? See also #435. It looks like this matching of `decimalType` with `DOUBLE` is pervasive, so I'm wondering if you had some reason for doing it that way originally.

kmicklas:
@tathougies @3noch What do you guys think? I think this is correct but I'm worried about breaking any existing users. I'd like to get this in the release if possible.

3noch:
Hm, for Postgres this seems easy since it couldn't have worked. For SQLite it's complicated. It appears we correctly *parse* `DECIMAL` as this type, but then we emit `DOUBLE` instead? So from the "roundtrip" perspective this is just an important bug to fix. But from a migration's perspective of "I created a table with DOUBLE and now it's expecting a different type" it's much more troubling.

Perhaps we can release the Postresql change immediately and spend more time thinking about SQLite?

kmicklas:
@3noch I'd like to revisit this since we're doing a number of breaking changes for the release anyway. In particular, the integer size changes will already potentially require users to rework their migrations.

Also note that, due to #431, migrations with `Double` on SQLite were already broken / not round-tripping before I fixed that with #435.

3noch:
Yeah SQLite seems fine. I'm more worried about PG. There is nothing in the types that will tell you that we suddenly expect a different SQL type.

3noch:
I think we should *at least* fix SQLite immediately. But for PG I fear that even a breaking API version won't be enough to prevent chaos.

kmicklas:
@3noch It's the other way around since Postgres totally rejects `DOUBLE`. It's SQLite that this would matter for, but these types don't mean anything in SQLite anyway and are purely informational. I'm struggling to think of any way that this could lead to data loss, corruption, or other ill effects besides surprise and having to change existing migrations.

3noch:
I'm sorry. I remembered it backwards when I posted. You're right. This seems safe.

PR #443, opened 2020-03-09
GHC 8.8 + Stackage LTS 16 Support

fosskers:
This PR massages the various `beam` libraries to keep it up-to-date with the ecosystem.

There is currently a compile error due to depending on too old a version of `haskell-src-exts`.
```
beam-migrate         > /home/colin/code/haskell/beam/beam-migrate/Database/Beam/Haskell/Syntax.hs:957:3: error:
beam-migrate         >     Not in scope: data constructor ‘Hs.ClassA’
beam-migrate         >     Module ‘Language.Haskell.Exts’ does not export ‘ClassA’.
beam-migrate         >     |  
beam-migrate         > 957 |   Hs.ClassA () (Hs.UnQual () (Hs.Ident () "BeamMigrateSqlBackend")) [ beTy ]
beam-migrate         >     |   ^^^^^^^^^
```
Notice that this constructor is present [in `1.21`](https://hackage.haskell.org/package/haskell-src-exts-1.21.1/docs/Language-Haskell-Exts-Syntax.html#v:ClassA) but is completely gone [in `1.22`](https://hackage.haskell.org/package/haskell-src-exts-1.22.0/docs/Language-Haskell-Exts-Syntax.html#t:Asst). The CHANGELOG says that they changed, but doesn't explain how to migrate. I hesitate to change this myself, since I don't know what it's for.

Once this error is fixed, the bounds for `haskell-src-exts` will have to be adjusted.


fosskers:
@tathougies do you have any insight into the `haskell-src-exts` change here?

fosskers:
Hi there, any thoughts?

fosskers:
Any updates?

tathougies:
Hey sorry I've been so absent recently. Yes, this change is good.

My thoughts on haskell-src-exts has mainly been to try and get rid of it. We only use it to generate haskell code, and we certainly don't need its full power. Someone I spoke with suggested using the template haskell library from GHC for code generation. I think that's fine, I just haven't gotten around to it yet.

fosskers:
Cool, so should we keep this paused or try to fix the breakage?

kmicklas:
@fosskers Can we just pin `haskell-src-exts` to 1.22 in `stack.yaml`? The base bounds on that version look like it should support GHC 8.8. Also note that some of these changes have been made in #463 now.

fosskers:
Oh, it looks like the version of `haskell-src-exts` available in the current Stackage LTS is indeed `1.22`. Let me revisit this on Monday.

fosskers:
I looked at this again, and if the goal is to get this back onto Stackage, then we can't down-pin to `haskell-src-exts-1.21`. The real solution is to drop that dep altogether, given that it's no longer maintained.

maerten:
If `ClassA` is the only issue, then perhaps `beam` can be migrated to `haskell-src-exts` 1.22, by updating any use of `ClassA` in the same way as `haskell-src-exts` has done:

```diff
-                return $ ClassA (fl l) c ts
+                return $ tyApps (fl l) (TyCon (fl l) c) ts
```

As found here: https://github.com/haskell-suite/haskell-src-exts/commit/c1a6ad50dad44cf93a8972b13cdf78673277cddd#diff-6b33b8281133b44bca064796018b9d96R209

`ClassA` seems to be only used here: https://github.com/haskell-beam/beam/blob/737b73c6ec1c6aac6386bf9592a02a91f34a9478/beam-migrate/Database/Beam/Haskell/Syntax.hs#L957

fosskers:
Got through it! Now just a few other minor things to fix...

fosskers:
Nix-based CI failing, seems to be pinned to old versions.

fosskers:
https://github.com/obsidiansystems/constraints-extras/issues/32 needs to be moved forward first.

fosskers:
`constraint-extras` got a bump, so this can finally move forward.

fosskers:
Can someone help with the Nix? Looking at the only `.nix` files in this repo (`shell.nix`), it's not clear what has to change in order for CI to accept the newer `haskell-src-exts`.

fosskers:
Otherwise, tests pass locally so this PR is good-to-go.

kmicklas:
@fosskers `haskell-src-exts` 1.23.0 is now supported as of #492. Can you remove changes related to that? Also, do you know if it's possible to support both the newer and older versions of `network` simultaneously? I feel like that must be possible but it's not obvious how to me.

fosskers:
I was on a project recently where it supported 2.8 and 3.x, I think without any changes. Let me check.

fosskers:
CI failing due to the `network` bounds.

fosskers:
The bounds work with Stackage `lts-14` and `lts-16`, but Nix still doesn't like it.

maerten:
Perhaps this fix would work @fosskers -> https://github.com/haskell-beam/beam/pull/517

PR #457, opened 2020-05-17
Add PolyKinds to modules deriving for Data.Tagged

srid:
This allows Tagged's phantom type to be of kind `Symbol`.

Context: https://haskell.zettel.page/047d9f68.html

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL.hs
+++ b/beam-core/Database/Beam/Backend/SQL.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE UndecidableInstances #-}
 module Database.Beam.Backend.SQL
   ( module Database.Beam.Backend.SQL.Row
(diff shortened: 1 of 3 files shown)


kmicklas:
I think this is probably fine but how do you feel about just adding kind signatures to the relevant instances? (`KindSignatures` is already enabled.) Sometimes `PolyKinds` ends up making things more polymorphic than you really intended.

srid:
So define instances for `Type` and `Symbol` separately? Yea, that sounds like a better approach.

kmicklas:
No, it should support any kind. You should be able to do something like:

```
instance (BeamBackend be, FromBackendRow be t) => FromBackendRow be (Tagged (tag :: k) t) where
```

srid:
> `tag :: k`

This seems to require `PolyKinds` again?

image

kmicklas:
Ah, you're right; I thought `KindSignatures` was sufficient for that. We shouldn't be adding individual kind-specific instances so this is good as-is.

srid:
Awesome, thanks! My first beam contribution. :-)

PR #458, opened 2020-05-17
Added Postgres specific || operator for concatenating two jsonb values

mputz86:
I am not sure if the operator naming is correct and if it is ok to introduce these PostgreSQL specific operators like this (not so much experienced in Haskell yet). But anyway, I try my luck and hope at least to get feedback on how to do it correctly :D.
Here is the official description of the `||` jsonb-concatenator operator as well as some more https://www.postgresql.org/docs/11/functions-json.html (maybe I can add some more ;) ).
Thank you.

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/PgSpecific.hs
+++ b/beam-postgres/Database/Beam/Postgres/PgSpecific.hs
@@ -32,7 +32,7 @@ module Database.Beam.Postgres.PgSpecific
 
   , (@>), (<@), (->#), (->$)
   , (->>#), (->>$), (#>), (#>>)
-  , (?), (?|), (?&)
+  , (?), (?|), (?&), (||.)
 
   , withoutKey, withoutIdx
   , withoutKeys
@@ -114,7 +114,7 @@ module Database.Beam.Postgres.PgSpecific
   )
 where
 
-import           Database.Beam hiding (char, double)
+import           Database.Beam hiding ((||.), char, double)
 import           Database.Beam.Backend.SQL
 import           Database.Beam.Migrate ( HasDefaultSqlDataType(..) )
 import           Database.Beam.Postgres.Syntax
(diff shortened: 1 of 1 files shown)


beam-postgres/Database/Beam/Postgres/PgSpecific.hs
```diff
@@ -960,6 +960,14 @@ QExpr a ?| QExpr b =
 QExpr a ?& QExpr b =
   QExpr (pgBinOp "?&" <$> a <*> b)
 
+-- | Postgres @||@ operator. Concatenates two jsonb values into a new jsonb value.
+(||.) ::
+  QGenExpr ctxt Pg.Postgres s (Pg.PgJSONB a) ->
```
kmicklas, line 965: nit: this arrow style is not the norm in Beam.

kmicklas (commented):
Thanks! It's definitely nice to support as many operators as we can.

beam-postgres/Database/Beam/Postgres/PgSpecific.hs
mputz86: Uh, totally right - fixed that (plus removing these `Pg.` prefixes`).

mputz86:
Also got some conflict with the `||.` from `Database.Beam`. So either the operators should be named different than in Postgres or (like now) hiding the one from `Database.Beam` and let the user import Postgres specific stuff qualified (like I did it :D 🤷‍♂️ ).

kmicklas:
Ah good point, I totally forgot that `||.` is already taken. Unless you can think of a better operator name that makes sense (I can't), then I'd suggest just giving it a name like `pgJsonbConcat`.

mputz86:
Then for me the question is "where" to put it in code:
- put it close (i.e. maybe at the end of) the `pgJsonbABC` things or
- keep it in the order of the Postgres doc (like obviously all the rest is right now) and put it after `(?&)` but before `withoutKey` :) 

Sry, too much discussion for such a small thing? :D

kmicklas:
I don't think it matters much but probably nice to keep the order in line with the Postgres docs where possible.

PR #472, opened 2020-06-09
Bump major version numbers

kmicklas:
Breaking changes in `beam-core`: #413 (unlikely to break anyone in practice).

Breaking changes in `beam-migrate`: #329.

Breaking changes in `beam-postgres`: #329, #347 and associated PRs.

Breaking changes in `beam-sqlite`: #435.

### DIFF
--- a/beam-core/beam-core.cabal
+++ b/beam-core/beam-core.cabal
@@ -2,7 +2,7 @@
 -- see http://haskell.org/cabal/users-guide/
 
 name:                beam-core
-version:             0.8.1.0
+version:             0.9.0.0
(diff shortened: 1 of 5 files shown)


3noch:
We need to update ChangeLogs too?

kmicklas:
Yep, we need to do that before the release.

kmicklas:
Hmm, the CI definitely didn't work on this. It succeeded without any output.

kmicklas:
> Hmm, the CI definitely didn't work on this. It succeeded without any output.

Or not: https://github.com/kmicklas/beam/runs/755025866

Something very weird is up with the `pull_request` version, even though the `push` one correctly failed.

PR #475, opened 2020-06-09
Fix domain in docs URLs

### LINKED ISSUE #471: Docs are down
I see a 404 at https://tathougies.github.io/beam

kmicklas:
Fixes #471.


3noch:
We should switch to https everywhere too I think.

kmicklas:
> We should switch to https everywhere too I think.

Good catch, done.

PR #476, opened 2020-06-09
Use newtype for BeamHasInsertOnConflict data family instances

kmicklas:
When I originally wrote these I believe I was unaware that these can be `newtype`s.

### DIFF
--- a/beam-sqlite/Database/Beam/Sqlite/Connection.hs
+++ b/beam-sqlite/Database/Beam/Sqlite/Connection.hs
@@ -409,9 +409,9 @@ runInsertReturningList (SqlInsert _ insertStmt_@(SqliteInsertSyntax nm _ _ _)) =
            return x
 
 instance Beam.BeamHasInsertOnConflict Sqlite where
-  data SqlConflictTarget Sqlite table = SqliteConflictTarget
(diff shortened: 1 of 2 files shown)


beam-sqlite/Database/Beam/Sqlite/Connection.hs
```diff
@@ -409,9 +409,9 @@ runInsertReturningList (SqlInsert _ insertStmt_@(SqliteInsertSyntax nm _ _ _)) =
            return x
 
 instance Beam.BeamHasInsertOnConflict Sqlite where
-  data SqlConflictTarget Sqlite table = SqliteConflictTarget
+  newtype SqlConflictTarget Sqlite table = SqliteConflictTarget
     { unSqliteConflictTarget :: table (QExpr Sqlite QInternal) -> SqliteSyntax }
-  data SqlConflictAction Sqlite table = SqliteConflictAction
+  newtype SqlConflictAction Sqlite table = SqliteConflictAction
```
3noch, line 414: Newtyping an existential works?

3noch: Oh I see now.

PR #487, opened 2020-06-13
Use tmp-postgres

### LINKED ISSUE #482: Use tmp-postgres for setting up test Postgres connections
This abstracts out the details of using `pg_ctl` and is likely faster.

3noch:
Fixes #482


3noch:
There are probably more optimizations we can do but this is a start.

3noch:
I've got to fix up release.nix

3noch:
Bummer. I'll have to patch `generic-monoid` to support ghc-8.4....

3noch:
Currently "blocked" on
  * https://github.com/luke-clifton/generic-monoid/pull/2
  * https://github.com/jfischoff/postgres-options/issues/3

There is probably a way to get things moving without these, but it probably also isn't worth the effort.

luke-clifton:
`generic-monoid` is updated and 0.1.0.1 is released on Hackage with the required changes.

3noch:
Thanks @luke-clifton!

kmicklas:
I've seen what looks like a non-determinism issue with Postgres startup in the tests a few times now: https://github.com/haskell-beam/beam/pull/507/checks?check_run_id=963783156

I wonder if this PR would help with that. (I've been using tmp-postgres in tests in a Nix build for a while now and seen nothing similar.)

tathougies:
There's an auto-save artifact in this commit that should be dropped before committing.

3noch:
We are a long way off from being able to commit this unless we drop support for GHC 8.4.

3noch:
@tathougies Actually I think I *deleted* the auto save artifact in this PR.

tathougies:
FWIW, my patch for GHC >8.8 compatibility dropped GHC 8.4 support. I think it's okay to drop support for older GHC.

3noch:
@tathougies @kmicklas Ok this finally passes CI. It includes a commit to drop GHC 8.4 but it does *not* include a commit to add GHC 8.10.

tathougies:
I think this is ready to merge. Can we do it?

PR #488, opened 2020-06-16
Add postgres extensions to database predicates

blombern:
Currently `verifySchema` doesn't pick up extensions, and thus fails if `CheckedDatabaseSettings` specifies any.

### DIFF
--- a/beam-postgres/test/Database/Beam/Postgres/Test/Migrate.hs
+++ b/beam-postgres/test/Database/Beam/Postgres/Test/Migrate.hs
@@ -3,6 +3,7 @@ module Database.Beam.Postgres.Test.Migrate where
 import Database.Beam
 import Database.Beam.Postgres
 import Database.Beam.Postgres.Migrate
+import Database.Beam.Postgres.PgCrypto (PgCrypto)
 import Database.Beam.Postgres.Test
 import Database.Beam.Migrate
 import Database.Beam.Migrate.Simple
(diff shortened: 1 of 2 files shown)


3noch:
Thanks! I'd like to see if we can add a test for this.

blombern:
I'll try to find a bit of time for it in the coming days 👍

beam-postgres/test/Database/Beam/Postgres/Test/Migrate.hs
```diff
@@ -70,3 +77,15 @@ charNoWidthVerification pgConn tyName charTy =
...
             VerificationFailed failures -> fail ("Verification failed: " ++ show failures)
+
+-- | Verifies that 'verifySchema' correctly checks enabled PgCrypto extension
+extensionVerification :: IO ByteString -> TestTree
+extensionVerification pgConn =
+    testCase "verifySchema correctly checks enabled PgCrypto extension" $
+      withTestPostgres "db_extension_pgcrypto" pgConn $ \conn ->
+        runBeamPostgres conn $ do
+          db <- executeMigration runNoReturn (CryptoDb <$> pgCreateExtension)
+          res <- verifySchema migrationBackend db
```
3noch, line 88: Seems like we should run `verifySchema` both before and after migration to ensure it's not a false negative?

3noch:
@kmicklas @tathougies Since I modified this I'd like one of you to review as well.

PR #492, opened 2020-07-11
Bump haskell-src-exts

roberth:


kmicklas:
I verified this locally, but I don't have permissions to push to the PR branch to update the nix. I'll put up a separate PR for that.

3noch:
Thanks @kmicklas. I forgot to come back to this.

Issue #494, opened 2020-07-27
haskellSchema will fail if any table in the schema has no PK


o1lo01ol1o:
This is an additional consequence of #447 but I didn't see that issue until after I had debugged `haskellSchema` far enough to understand that `TableHasPrimaryKey "table" []` is not satisfiable.

kmicklas:
@o1lo01ol1o I believe this is fixed now by #495, please reopen if I'm mistaken. (Unfortunately we don't have tests for this yet.)

PR #495, opened 2020-07-28
Suppress creation of primary key constraints for tables with no primary key

### LINKED ISSUE #447: TableHasPrimaryKey check is unsatisfiable when PK is null.
The workaround for tables without a primary key is to define

```haskell
    data PrimaryKey TableT f = TableHasNoPK
```

however `beam-migrate` will collect a `TableHasPrimaryKey "table" []` check, which is not satisfiable.

The `TableCheck` is defined to be `... -> SomeDatabasePredicate`, so it has to return something.

- One way is to make `TableCheck` return `Maybe SomeDatabasePredicate`.
- Another way is to add `TrueDatabasePreciate`, which are filtered out in `verifySchema` and alike functions.

kmicklas:
Fixes #447 and possibly #494.


3noch:
Awesome!

kmicklas:
@3noch I added a test for SQLite (which would fail previously). I'd like to add a test for Postgres eventually too but it shouldn't block merging this PR now.

Issue #496, opened 2020-07-29
Clean up integer size handling


kmicklas:
#329 greatly improved support for different integer sizes, however this introduced more issues and confusion (for example #493).

These are the steps I think we should take to rectify the situation:

 * Make all implementation-specific functions (e.g. `pgUnnestArrayWithOrdinality`) use correct explicitly sized integers (in that case `Int64` for `bigint`).
 * Create a type family for the cardinality returned by `COUNT` and the like, and use this to correct combinators that return `Int` like `countAll_`. Alternatively, make these polymorphic over any `Integral` like `count_`.
 * Remove the `Int` instances and add a custom type error directing the user to use explicitly sized integers.

Ideally we don't support `Int` anywhere since it's ambiguous in Haskell and database interop isn't something you generally want to leave to chance.

I'd like to get this in before the next release since it's a fairly large breaking change.

3noch:
Oof great points. Yes this seems very important.

PR #497, opened 2020-07-29
Use explicitly sized integer types for Postgres-specific functions

kmicklas:
This is the first step of #496.

Relevant documentation:

 * `UNNEST ... WITH ORDINALITY`: https://www.postgresql.org/docs/12/queries-table-expressions.html
 * JSON functions: https://www.postgresql.org/docs/current/functions-json.html
 * pgcrypto: https://www.postgresql.org/docs/current/pgcrypto.html

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/PgSpecific.hs
+++ b/beam-postgres/Database/Beam/Postgres/PgSpecific.hs
@@ -134,6 +134,7 @@ import qualified Data.ByteString.Char8 as BC
 import qualified Data.ByteString.Lazy as BL
 import           Data.Foldable
 import           Data.Hashable
+import           Data.Int
 import qualified Data.List.NonEmpty as NE
 import           Data.Proxy
 import           Data.Scientific (Scientific, formatScientific, FPFormat(Fixed))
@@ -888,7 +889,7 @@ QExpr a <@ QExpr b =
 -- See '(->$)' for the corresponding operator for object access.
 (->#) :: IsPgJSON json
       => QGenExpr ctxt Postgres s (json a)
-      -> QGenExpr ctxt Postgres s Int
+      -> QGenExpr ctxt Postgres s Int32
       -> QGenExpr ctxt Postgres s (json b)
 QExpr a -># QExpr b =
   QExpr (pgBinOp "->" <$> a <*> b)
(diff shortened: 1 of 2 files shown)


beam-postgres/Database/Beam/Postgres/PgSpecific.hs
```diff
@@ -1409,14 +1410,14 @@ pgUnnestArray (QExpr q) =
...
-data PgUnnestArrayWithOrdinalityTbl a f = PgUnnestArrayWithOrdinalityTbl (C f Int) (C f a)
+data PgUnnestArrayWithOrdinalityTbl a f = PgUnnestArrayWithOrdinalityTbl (C f Int64) (C f a)
   deriving Generic
 instance Beamable (PgUnnestArrayWithOrdinalityTbl a)
 
 -- | Introduce each element of the array as a row, along with the
 -- element's index
 pgUnnestArrayWithOrdinality :: QExpr Postgres s (V.Vector a)
-                            -> Q Postgres db s (QExpr Postgres s Int, QExpr Postgres s a)
+                            -> Q Postgres db s (QExpr Postgres s Int64, QExpr Postgres s a)
```
3noch, line 1420: How are deciding which Int type to use?

kmicklas: Postgres documentation: https://www.postgresql.org/docs/12/queries-table-expressions.html

3noch: Oh I see it's the `bigint` here in the docs for `WITH ORDINALITY`... I wish there were a more obvious way to connect this choice with the docs.

3noch: I'm normally not a big fan of type aliases but perhaps it makes sense to define `type PgBigInt = Int64` to make it more obvious. But even then we have to go digging through docs again to see where the bigint shows up.

kmicklas: Hmm, I think I'd rather not use type aliases. Besides all the usual reasons, it seems a bit more confusing to have something like `type PgInt = Int32`.

3noch: Yeah I'm not really a fan either. It's just that there's not a very mechanical way to verify these kinds of choices strewn throughout the codebase. Perhaps that's just par for the course.

PR #500, opened 2020-08-01
Use Integral for SQL Int values in beam-core

kmicklas:
This should remove all SQL-internal uses of `Int` in beam-core (#496). Rather than introducing a type family for the cardinality type for a backend, I extended the existing pattern of using `Integral`, even though it's in general less safe.


3noch:
I take it the idea is that anywhere we have `Integral a => a`, the conversion to `a` happens "client-side" and the datatype encoding of the DB type is still well-defined?

3noch:
Also, should this count as a breaking change since users may need to add type annotations or `as_` clauses as you've done.

kmicklas:
Yeah, it is a breaking change. I'd like to get this and related PRs in before the release.

PR #503, opened 2020-08-04
Add Postgres extension definition for uuid-ossp

kmicklas:
As of #488 extensions are picked up by `backendGetDbConstraints`, so it becomes more useful to have them predefined.

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Extensions/Internal.hs
+++ b/beam-postgres/Database/Beam/Postgres/Extensions/Internal.hs
@@ -1,0 +1,17 @@
+module Database.Beam.Postgres.Extensions.Internal where
+
+import Data.Text (Text)
+
+import Database.Beam
+import Database.Beam.Backend.SQL
+import Database.Beam.Postgres
+
+type PgExpr ctxt s = QGenExpr ctxt Postgres s
+
+type family LiftPg ctxt s fn where
+  LiftPg ctxt s (Maybe a -> b) = Maybe (PgExpr ctxt s a) -> LiftPg ctxt s b
+  LiftPg ctxt s (a -> b) = PgExpr ctxt s a -> LiftPg ctxt s b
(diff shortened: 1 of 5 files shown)


beam-postgres/Database/Beam/Postgres/Extensions.hs
```diff
@@ -195,3 +195,12 @@ pgDropExtensionProvider =
...
                            ("Unload the postgres extension " <> ext) 1)
+
+-- *** Built-in extensions
+
+-- | Data type representing definitions contained in the @plpgsql@ extension
+data PlPgSql = PlPgSql
+
+instance IsPgExtension PlPgSql where
+  pgExtensionName Proxy = "plpgsql"
+  pgExtensionBuild = PlPgSql
```
3noch, line 206: Are there any tests for this we can add?

3noch: I realize there are no methods to this extension... which begs the question: Why have this at all?

kmicklas: I needed this because I'm using a custom version of `verifySchema` which checks that there are no "extra" predicates in addition to that no expected predicates are missing. I still need to upstream that so happy to remove it from this PR if you think it's not useful.

3noch: Hm, I don't have a strong opinion if you have plans to upstream that stuff anyway. The disadvantage of adding it now is that changing it later means major version bump whereas adding it all at once later is a minor version bump.

kmicklas: I removed `plpgsql` for now. I'll put that in a separate PR when I upstream the two-way verification support.

kmicklas: For reference the feature I was talking about is now in #522.

PR #505, opened 2020-08-08
Remove instances for ambiguously sized integers

### LINKED ISSUE #496: Clean up integer size handling

kmicklas:
Fixes #496. It's possible I missed some, but all instances for `Int` and `Word` should now be replaced with a helpful custom type error.

### DIFF
--- a/beam-core/Database/Beam/Backend/Internal/Compat.hs
+++ b/beam-core/Database/Beam/Backend/Internal/Compat.hs
@@ -1,0 +1,16 @@
+-- | This module contains utilities that backend writers can use to assist with
+-- compatibility and breaking API changes.
+--
+-- Users should not need anything from this module.
+module Database.Beam.Backend.Internal.Compat where
+
+import GHC.TypeLits
+
+-- | A type error directing the user to use an explicitly sized integers,
+-- instead of 'Int' or 'Word'.
+type PreferExplicitSize implicit explicit =
+  'Text "The size of " ':<>:
+  'ShowType implicit ':<>:
+  'Text " is machine-dependent. Use an explicitly sized integer such as " ':<>:
+  'ShowType explicit ':<>:
+  'Text " instead."
--- a/beam-core/Database/Beam/Backend/SQL/AST.hs
+++ b/beam-core/Database/Beam/Backend/SQL/AST.hs
@@ -6,6 +6,7 @@ module Database.Beam.Backend.SQL.AST where
 
 import Prelude hiding (Ordering)
 
+import Database.Beam.Backend.Internal.Compat
 import Database.Beam.Backend.SQL.SQL92
 import Database.Beam.Backend.SQL.SQL99
 import Database.Beam.Backend.SQL.SQL2003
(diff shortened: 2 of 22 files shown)


kmicklas:
> What if instead of doing this right away we simply deleted the instances for now and then introduced these instances in the very next version as a way to let people migrate? I.e. they can write orphans for one version and then they have to fully migrate?

I'm not exactly following what you mean by this. Would that be just not adding the explicit type errors so that it's still possible to add the instances as orphans (which we could potentially provide)? That would be more flexible but the downside is the errors would be way more confusing. With the custom errors, the message is pretty clear: replace `Int` in your schema with something else.

> As an alternative we could add a newtype "UnsafeMachineInt" or something as a helper to people trying to migrate.

I think changing your types to `UnsafeMachineInt` and inserting all the necessary coercions is no harder than changing to `Int32`.

3noch:
Yeah `UnsafeMachineInt` isn't any better. The problem I'm imagining is that people use schema types in many more places than just their schema. Those types end up on web frontends often. So now they have to modify not just their schema but literally every part of their program. It's a pretty hefty migration path. At the very least a newtype is coercible to `Int` so maybe there's some hope that a `Coercible a Int` constraint could help them avoid the ripple effects. I agree though that the band aid must at some point come off and then the migration simply must happen. I'm just trying to think of ways we could make it easier on people.

One way I've thought about this kind of thing in the past: If I have important bug fixes in the release queue and big API breaking changes in the queue, try to find a way to get the bug fixes out in release A and then the API breakage in release B. That way people can at least get the bug fixes without paying the hefty breakage. Of course our next release is API breaking anyway, but this particular change is *really* breaking. I'm not against moving forward with this. I just want to make sure we've at least discussed the implications.

kmicklas:
Since it's been well over a year since any release at all, and we already have breaking changes, I think we should try to get all planned breaking changes into this release. Getting that right, so we're not immediately planning another major release, has been my goal so far. (There are a number of non-breaking enhancement PRs that I would like to clean up and merge, but those can go in a minor release shortly after.)

I wish this wasn't such a large change but I'm hoping the actual effort of migration will be fairly minimal and mechanical. By comparison, I think beam-core 0.7 to 0.8 was much more intensive.

3noch:
Fair enough. I'll defer to your judgment.

PR #512, opened 2020-08-20
Add tentative ChangeLog entries

kmicklas:
Rendered:

 * [beam-core](https://github.com/kmicklas/beam/blob/release-notes/beam-core/ChangeLog.md)
 * [beam-migrate](https://github.com/kmicklas/beam/blob/release-notes/beam-migrate/ChangeLog.md)
 * [beam-postgres](https://github.com/kmicklas/beam/blob/release-notes/beam-postgres/ChangeLog.md)
 * [beam-sqlite](https://github.com/kmicklas/beam/blob/release-notes/beam-sqlite/ChangeLog.md)

PR #514, opened 2020-08-28
Fix docs

tathougies:
This fixes docs enough to work on my machine, which is on latest NixOs 20.03. There are some things you need to do to build the docs (like have a postgres server running on localhost:5432). Disabled beam-mysql and beam-mssql docs because they need to be updated.


tathougies:
Oh I will say that we should probably merge this before #512 , and that I still need to work on the `ON CONFLICT` sections.

nixpkgs/mkshell.nix
```diff
@@ -0,0 +1,24 @@
+{ nixpkgs ? import <nixpkgs> {}, ghc ? nixpkgs.haskell.packages.ghc881.ghc, docs ? false }:
```
kmicklas, line 1: Can we have this default to a pinned nixpkgs version which is shared with `release.nix` for CI? The version there is 20.03 so it would likely work.

tathougies: Yes, let's pin it.

tathougies: If I pin it, will it cause issues with the GitHub Ci?

3noch: It shouldn't.

3noch: We should pull the pin out of release.nix and use it here and there so they always use the same pin though.

3noch: I guess that isn't strictly necessary, but if we actually can't use the same pin that would be a bit unfortunate.

docs.nix
```diff
@@ -0,0 +1 @@
+import ./nixpkgs/mkshell.nix { docs = true; nixpkgs = import ./nixpkgs/pinned.nix; }
```
3noch, line 1: Looks like this pinned file didn't make it into the commit.

tathougies:
Everything should be there now.

tathougies:
Next step is to get rid of the stack based build system for docs, and then we should be able to have GitHub build them all.

docs/user-guide/manipulation/insert.md
```diff
@@ -232,3 +232,109 @@ runInsert $ insert (playlist chinookDb) $
...
+
+runInsert $
+  insertOnConflict (customer chinookDb) (insertValues [newCustomer])
+     anyConflict
+     onConflictDoNothing
+```
+
+### Acting only on certain conflicts
+
+SOmetimes you only want to perform an action if a certain constraint is violated. IF the conflicting
```
kmicklas, line 290: typos: SOmetimes, IF

docs/user-guide/manipulation/insert.md
```diff
@@ -232,3 +232,109 @@ runInsert $ insert (playlist chinookDb) $
...
+  newCustomer = Customer 42 "John" "Doe" Nothing (Address (Just "Street") (Just "City") (Just "State") Nothing Nothing) Nothing Nothing "john.doe@johndoe.com" nothing_
+
+runInsert $
+  insertOnConflict (customer chinookDb) (insertValues [newCustomer])
+    (conflictingFields (\tbl -> primaryKey tbl))
+    (onConflictUpdateSet (\fields oldValues -> fields <-. val_ newCustomer))
+```
+
+!!! tip "Tip"
+    To specify a conflict on the primary keys, use `conflictingField primaryKey`.
```
kmicklas, line 307: `conflictingFields`

nixpkgs/mkshell.nix
kmicklas: I want to make the docs build in CI, so ideally they can share the version. (We can of course update it if need be.)

docs/beam.yaml
```diff
@@ -63,49 +63,50 @@ backends:
...
-    extra-packages:
-      - mysql-simple
-    extra-imports:
-      - Database.MySQL.Simple
-      - "qualified Database.MySQL.Base as Base"
-    backend-extra: >-
-      docsWithTransaction = withTransaction
-    supports:
-      - utf8
+#  beam-mysql:
```
3noch, line 66: Is there a reason to keep these commented?

docs/user-guide/manipulation/insert.md
```diff
@@ -232,3 +232,109 @@ runInsert $ insert (playlist chinookDb) $
...
+
+* `onConflictDoNothing` - this cancels the insertion
+* `onConflictUpdateSet` - sets fields to new values based on the current values
+* `onConflictUpdateSetWhere` - sets fields to new values if a particular condition holds
+
+
+### Acting on any conflict
+
+A common use case of `ON CONFLICT` is to *upsert* rows into a database. *Upsertion* refers to only
+inserting a row if another conflicting row does not already exist. For example, if you have a new
```
3noch, line 272: I think upsert refers to "Updating *or* inserting*". In this case you're inserting without updating.

3noch: I guess that's just being super pedantic and probably different people don't agree on the specifics.

nixpkgs/pinned.nix
```diff
@@ -0,0 +1,6 @@
+import ((import <nixpkgs> {}).fetchFromGitHub {
```
3noch, line 1: We should just use `builtins.fetchGit` instead to get rid of this impurity:

```import (builtins.fetchGit {
  url = https://github.com/NixOS/nixpkgs.git;
  rev = "5272327b81ed355bbed5659b8d303cf2979b6953";
  sha256 = "0182ys095dfx02vl2a20j1hz92dx3mfgz2a6fhn31bqlp1wa8hlq";
}) {}

3noch: I removed the `{}` at the end because it's customary to allow the caller to pass in arguments and it better matches the `import <nixpkgs> {}` idiom when you switch to `import ./mypin.nix {}`

docs/beam.yaml
tathougies: Yeah, it doesn't work when I uncomment them :P 

I'll fix them, but I think I want to move the docs generation scripts into a new repository so there's no circular dependency on alternative backends...

docs/user-guide/manipulation/insert.md
tathougies: Yeah i'll change that..

tathougies:
I'm going to merge and then make a separate MR for the pinning stuff

PR #515, opened 2020-08-28
Fix warnings on latest GHC 8.8.1

tathougies:
I also think it's time we dropped support for ancient Preludes. I started removing some useless CPP macros. Hoping to get rid of all of them.

### DIFF
--- a/beam-migrate-cli/beam-migrate-cli.cabal
+++ b/beam-migrate-cli/beam-migrate-cli.cabal
@@ -47,7 +47,7 @@ executable beam-migrate
                        network              >=2.6      && <2.9,
                        yaml                 >=0.8      && <0.12,
                        uuid                 >=1.3      && <1.4,
-                       hint                    >=0.6      && <0.10,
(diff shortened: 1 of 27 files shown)


beam-migrate-cli/beam-migrate-cli.cabal
```diff
@@ -47,7 +47,7 @@ executable beam-migrate
                        network              >=2.6      && <2.9,
                        yaml                 >=0.8      && <0.12,
                        uuid                 >=1.3      && <1.4,
-                       hint                 >=0.6      && <0.10,
+                       hint                    >=0.6      && <0.10,
```
kmicklas, line 50: Did you mean to commit this?

tathougies: no

tathougies: BTW, I use haskell-mode's standard import formatter. Not sure how to reproduce it. I'm open to adopting a styling tool, but don't know much about them.

PR #517, opened 2020-08-31
Fix nixpkgs pinning

tathougies:
This is per @3noch's suggestion

### DIFF
--- a/nixpkgs/pinned.nix
+++ b/nixpkgs/pinned.nix
@@ -1,6 +1,5 @@
-import ((import <nixpkgs> {}).fetchFromGitHub {
-    owner = "NixOS";
-    repo = "nixpkgs";
+import (builtins.fetchGit {
+    url = https://github.com/NixOS/nixpkgs.git;
     rev = "5272327b81ed355bbed5659b8d303cf2979b6953";
     sha256 = "0182ys095dfx02vl2a20j1hz92dx3mfgz2a6fhn31bqlp1wa8hlq";
-  }) {}
+})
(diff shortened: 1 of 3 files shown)


nixpkgs/pinned.nix
```diff
@@ -1,6 +1,5 @@
-import ((import <nixpkgs> {}).fetchFromGitHub {
-    owner = "NixOS";
-    repo = "nixpkgs";
+import (builtins.fetchGit {
+    url = https://github.com/NixOS/nixpkgs.git;
     rev = "5272327b81ed355bbed5659b8d303cf2979b6953";
     sha256 = "0182ys095dfx02vl2a20j1hz92dx3mfgz2a6fhn31bqlp1wa8hlq";
```
3noch, line 4: `fetchGit` doesn't take a `sha256` argument.

kmicklas: Oh I didn't realize this changed to `fetchGit`, I thought that was already broken. Relatedly, I think we should use `fetchTarball` for nixpkgs so we don't fetch unneeded history.

3noch: Good point.

tathougies: How would that work?

3noch: This is what it looks like for latest commit of 20.03: 

```nix
import (builtins.fetchTarball {
  url = https://github.com/NixOS/nixpkgs-channels/archive/ff6a070b4efdb7c4a37efef42d24a226565342d4.tar.gz;
  sha256 = "1xw6xqv4wr6ljnlp401bynw9hrcirq4jf59x1zyiyxf146gy9pg6";
})
```

I used `nix-prefetch-url --unpack <url>` to get the hash.

Issue #372, opened 2019-04-18
SQLite Beam.runSelectReturningList causes BadRowReadErrors when order of columns in record doesn't match order in database


moll:
Hey again!

That's an interesting and rather dangerous bug. If the model's column declaration on the Haskell side doesn't match the column order in the SQLite table, the `RETURNING` emulation of https://github.com/tathougies/beam/blob/d87120b58373df53f075d92ce12037a98ca709ab/beam-sqlite/Database/Beam/Sqlite/Connection.hs ends up creating records with wrong values and may even throw `BeamRowReadError` if the types don't match.

PR #519, opened 2020-09-14
Fix beam-sqlite runInsertReturningList

tathougies:
runInsertReturningList on Sqlite does a bit of a dance to get the results out. The temporary table has to be created using a `SELECT *` so that type information which `beam-core` does not have available is propagated to the new temporary table. However, this can cause deserialization issues if the fields in SQLite are not in the same order as in Haskell. This is fixed by ensuring we `SELECT` from the temporary table using an explicit projection.

### DIFF
--- a/beam-sqlite/ChangeLog.md
+++ b/beam-sqlite/ChangeLog.md
@@ -1,3 +1,7 @@
+# 0.5.0.0
+
+Fix `runInsertReturningList` for when the database column order and beam column order disagree.
+
 # 0.3.2.0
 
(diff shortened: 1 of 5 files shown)


kmicklas:
Can you add a test for this? We have SQLite selection tests now in `Database.Beam.Sqlite.Test.Select`.

kmicklas:
Also for reference this fixes #372.

3noch (approved):

3noch:
Now we just need a changelog entry for this since it snuck in last minute :D

Issue #474, opened 2020-06-09
beam-postgres should not use cursors by default


lpsmith:
While fighting a production fire yesterday involving an slow SQL query,  I discovered that beam-postgres uses cursors for literally every user query.  Cursors are occasionally useful in specialized cases,  but probably not worth it in typical use cases.   This definitely should not be default behavior.

Disadvantages of cursors:

1.  More round trips,  more overhead,  greater latency

2.  interacts badly with `log_min_duration_statement`,  making performance troubleshooting much more difficult.   Here's an anonymized excerpt of the actual log from yesterday with `log_min_duration_statement = 1000`.

```
Jun 08 15:10:23 postgresql-start[16980]: LOG:  duration: 2388.583 ms  statement: FETCH FORWARD 256 FROM temp26
Jun 08 15:10:23 postgresql-start[16980]: LOG:  duration: 3247.626 ms  statement: FETCH FORWARD 256 FROM temp6
Jun 08 15:10:24 postgresql-start[16980]: LOG:  duration: 2848.849 ms  statement: FETCH FORWARD 256 FROM temp26
Jun 08 15:10:25 postgresql-start[16980]: LOG:  duration: 4216.754 ms  statement: FETCH FORWARD 256 FROM temp26
```

Without logging the corresponding `DECLARE`  cursor statement, which is almost always very fast,  none of these log lines are actionable.  So we had to resort to `log_min_duration_statement = 0` and finding long-running fetch forward statements and then find corresponding declare cursor statements.    Furthermore,  at least with the version of postgres we are currently using for this project (`9.6`, I believe) even with `log_min_duration_statement = 0` we weren't always getting the declare queries in the log,  and I experimented with `log_statement` and `log_duration` with no luck.  (Unfortunately `log_duration` is almost entirely useless in production, as it logs the statements and durations in separate log lines in a racy way, so the correspondence between durations and statements is frequently ambiguous.  Using `log_min_duration_statement` in conjunction with `log_statement` results in the same behavior,  and not (as I was hoping) logging every statement twice,  one with a duration.  But I digress, as this is likely a postgres bug)

There are probably more disadvantages, but these are what came to mind.

Use cases for cursors:

[...]

PR #520, opened 2020-09-16
runReturningOne/List fetches all rows at once

rimmington:
Minimal fix for #474.

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Connection.hs
+++ b/beam-postgres/Database/Beam/Postgres/Connection.hs
@@ -184,7 +184,8 @@ withPgDebug dbg conn (Pg action) =
       step (PgLiftIO io next) = io >>= next
       step (PgLiftWithHandle withConn next) = withConn conn >>= next
       step (PgFetchNext next) = next Nothing
-      step (PgRunReturning (PgCommandSyntax PgCommandTypeQuery syntax)
+      step (PgRunReturning CursorBatching
+                           (PgCommandSyntax PgCommandTypeQuery syntax)
                            (mkProcess :: Pg (Maybe x) -> Pg a')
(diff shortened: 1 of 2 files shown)


tathougies:
Can you please add some postgres tests using cursors? This looks good to me otherwise.

beam-postgres/Database/Beam/Postgres/Connection.hs
```diff
@@ -201,7 +202,23 @@ withPgDebug dbg conn (Pg action) =
...
+      step (PgRunReturning _ (PgCommandSyntax _ syntax) mkProcess next) =
+        do query <- pgRenderSyntax conn syntax
+           dbg (T.unpack (decodeUtf8 query))
+           _ <- Pg.execute_ conn (Pg.Query query)
+
+           let Pg process = mkProcess (Pg (liftF (PgFetchNext id)))
+           runF process next stepReturningNone
+
+      renderExecReturningList :: (FromBackendRow Postgres x) => _ -> PgSyntax -> (Pg (Maybe x) -> Pg a') -> _ -> _
+      renderExecReturningList errMsg syntax mkProcess next =
```
tathougies, line 221: If `errMsg` isn't used, why is it here? I would think that any `*returningList` function can return an empty list without issue.

rimmington:
I added a super basic regression test for `runReturningMany`. I'm struggling to think of a test case that could differentiate between the behaviour of `runReturningMany` and `runReturningList`; I think the only differences should be memory usage and latency?

3noch (approved):

tathougies:
No need to differentiate between behaviors. Just want to make sure we are set up to catch any regressions in `runReturningMany` in CI.

PR #525, opened 2020-10-21
Add beam-postgres ChangeLog entry for #520

kmicklas:
Unfortunately this note won't be in the official release distribution which includes #520, but adding it to master for posterity.


3noch:
Doh! Good catch.

PR #526, opened 2020-10-28
Add MonadBaseControl and MonadBase to beam-sqlite and beam-postgres

tathougies:
I think this is pretty straightforwards, and a worthwhile change.


beam-postgres/Database/Beam/Postgres/#Connection.hs#
```diff
@@ -0,0 +1,394 @@
+{-# OPTIONS_GHC -fno-warn-orphans -fno-warn-partial-type-signatures #-}
```
3noch, line 1: Perhaps we need a .gitignore for this kind of file.

3noch:
This seems generally fine, although I'm a bit grumpy that `Pg` has even a `MonadIO` instance at all. In order to write a safe transaction monad you have to wrap `Pg` and remove all lifting to `IO` and even `MonadCatch`. It essentially needs to be like `STM` and can't be a transformer. But that's not the point of `Pg` really so this is seems fine.

beam-postgres/Database/Beam/Postgres/#Connection.hs#
tathougies: Yeah...

kmicklas:
`runBeamPostgres` doesn't create a transaction so I don't think of `Pg` as a transaction monad.

beam-sqlite/beam-sqlite.cabal
```diff
@@ -1,5 +1,5 @@
 name:                beam-sqlite
-version:             0.5.0.0
+version:             0.6.0.0
```
kmicklas, line 2: Why do we need to bump the major version number for this?

tathougies: Habit after a new release. I suppose the changes only mandate 0.5.1.0

3noch: `Database.Beam.Postgres.Connection` exports `PgF` though so I think it is backward breaking...

3noch:
@kmicklas Right. We just need to make sure we never let it become more than what it is. I have PTSD from another library of which I won't speak...

beam-sqlite/beam-sqlite.cabal
3noch: Oh but `Database.Beam.Postgres.Connection` is hidden and it doesn't look like `PgF(..)` gets re-exported anywhere. It would help if these module names made hiddenness a bit clearer.

tathougies:
I am open to suggestions on how to arrange internal / user-facing modules. If anyone has experience, let's discuss that.

3noch:
@tathougies We can discuss that but the only thing blocking *this* PR is the extraneous file and, possibly, a gitignore entry preventing it.

3noch (approved):

3noch:
There's already a gitignore rule that covers the extra file.

Issue #544, opened 2021-01-24
Weird non-deterministic errors with postgres and jsonb in certain rare scenarios. 


reite:
I have a web application using beam that has been running on beam-core 0.6.0.0 and beam-postgres for almost 2 years without issue. I recently upgraded to beam-core 0.9.0.0 and started rarely getting a decodeUtf8 error on certain requests. After a lot of debugging I have narrowed it down to queries on JSONB columns in some cases will return data from different columns than those queried. I am using Postgres 12

I have been able to reproduce the errors on my database, but it only happens in very few scenarios so I am afraid it will be difficult to create a minimal example. In my database I have a JSONB column in a certain row that contains data like this (some data removed):

```
{
    "-Kw06qjdIOYmTC8StGdj": {
        "data": {
            "-Kw07GBzKMX7DwsfDfYS": {
                "fields": {
                    "invoice_no": "xxxx",
                    "invoice_curr": "xxx",
                    "invoice_amount": "xxxx"
                }
            },
            "-L25nTRq7BzgHeuRD7lT": {
                "files": [
                    {
                        "name": "Invoice xxxx.pdf",
                        "path": "xxx/xxx/Invoice xxxx.pdf"
                    }
                ]
            }
        },
        "completed": {
            "by": "VBKADPAK5SZv2wX8FBg1UBIUIUG3",
            "when": "2018-01-31T14:34:34.519Z"
        }
    },
}
```

I can query this column by running this query with beam:

```
main = do
  connection <- connectPostgreSQL "..."

  pgjsonResult <- runBeamPostgresDebug putStrLn connection $ runSelectReturningOne $ select $ do
                    entity <-  all_ (getTable1 myDatabase)
                    guard_ (getEntityId entity ==. val_ "SomeId"))
                    return $ getPgJSONBField entity 

  print pgjsonResult 
```

When I run this query I get the expected result every time. The weirdness appears when I run a second query on another table that contains PgJSONB columns:

```
main = do
  connection <- connectPostgreSQL "..."

  pgjsonResult <- runBeamPostgresDebug putStrLn connection $ runSelectReturningOne $ select $ do
                    entity <-  all_ (getTable1 myDatabase)
                    guard_ (getEntityId entity ==. val_ "SomeId"))
                    return $ getPgJSONBField entity 

  runBeamPostgresDebug putStrLn connection $ runSelectReturningOne $ select $ do
        entity2 <- all_ (getTable2 glomoSqlDb)
[...]

PR #546, opened 2021-01-28
Copy Postgres result data when reading rows

kmicklas:
Tentative fix for #544.

We currently [manually free](https://github.com/haskell-beam/beam/blob/4d30ed45b16c41be26cb6301e9033c99379bf2d9/beam-postgres/Database/Beam/Postgres/Connection.hs#L232) the Postgres result object, and it seems that at least for `PgJSONB`, but likely also `PgJSON`, `Text, `ByteString`, etc. it's possible a slice of the underlying bytestring can outlive the result and cause a use-after-free bug.

`postgresql-simple` apparently also [copies](http://hackage.haskell.org/package/postgresql-simple-0.6.4/docs/Database-PostgreSQL-Simple-FromField.html#v:fromField) the data instead of letting GC handle the `Result`, which would be the other possible fix.


kmicklas:
@tathougies Any theories why no one detected this earlier, or why it would only manifest with the new non-cursor implementation?

PR #552, opened 2021-02-23
Order log entries when verifying migration status

rimmington:
If Postgres decides to return migration log entries in an unexpected order, `bringUpToDateWithHooks` will loop forever. I'm not sure _why_ it loops, but the query should have an `order by` anyway.


tathougies:
This is a good catch. It is odd that Postgres would even return them in any other order, since it's supposed to be the primary key. Nevertheless, without an order by, no RDBMS has an obligation.

rimmington:
@tathougies @kmicklas Anything I can do to speed this along?

kmicklas:
Sorry for the delay!

Issue #553, opened 2021-03-05
beam-sqlite fails to encode/decode some UTCTime values


thomasjm:
When I try to store and retrieve a `UTCTime` representing the year `1`, I get a decoding error.

```haskell
time = UTCTime (fromGregorian 1 0 0) (secondsToDiffTime 0)
```

The error is `BeamRowReadError {brreColumn = Just 5, brreError = ColumnTypeMismatch {ctmHaskellType = "UTCTime", ctmSQLType = "TEXT", ctmMessage = "conversion failed: couldn't parse UTCTime field: Failed reading: year must consist of at least 4 digits"}}`. 

I think this is because the time is getting encoded like this: `1-01-01T00:00:00`.

I've added a failing test to `beam-sqlite`, I'll open a separate PR with it.

tathougies:
`beam-sqlite` re-uses the parsers from sqlite-simple. My suggestion for this particular use case, since this is ultimately a non-standard format for dates, is to define

```
newtype MyTime = MyTime UTCTime
  deriving (Generic)
deriving newtype instance HasSqlValueSyntax syntax UTCTime => HasSqlValueSyntax syntax MyTime
instance FromBackendRow be Text => FromBackendRow be MyTime where
   fromBackendRow = mySpecialFunctionToParseTimes =<< fromBackendRow
mySpecialFunctionToParseTimes :: MonadFail m => Text -> m MyTime
```

As it is, I don't want to special-case parsing in Beam over whatever sqlite-simple does. IMO, this follows the principle of least surprise for users coming to beam from these libraries. Especially with dates, 01 could mean 2001 or 1901.

thomasjm:
> `beam-sqlite` re-uses the parsers from sqlite-simple

I just dug into this a little more, I think the issue is that the *encoding* part is done by `beam-sqlite` in a weird way:

https://github.com/haskell-beam/beam/blob/master/beam-sqlite/Database/Beam/Sqlite/Syntax.hs#L951-L953

In the `HasSqlValueSyntax` instance, `beam-sqlite` is using its own date format string rather that `sqlite-simple`'s `ToField` instance.

I updated my PR to use the `toField` function and it fixes the failing test: https://github.com/haskell-beam/beam/pull/554/commits/877a1b51846609b9b7a4509a7a55ba2b2ae069ce. What do you think? I think a couple other types like `Day` would also benefit from the same change.

kmicklas:
[...]

PR #554, opened 2021-03-05
Add failing test for UTCTime years

thomasjm:
For issue #553.


thomasjm:
@kmicklas if there's a major release coming, any chance we could get this in?

PR #561, opened 2021-04-19
Fix several typos

alexfmpe:
Found these way back when first reading the user guide then forgot to PR for two years

### DIFF
--- a/beam-migrate/Database/Beam/Migrate.hs
+++ b/beam-migrate/Database/Beam/Migrate.hs
@@ -16,10 +16,9 @@
 -- database entities, a checked database (represented by the
 -- 'CheckedDatabaseSettings' type) consists of a set of database entities along
 -- with a set of /predicates/ (represented by types implementing
--- 'DatabasePredicate').The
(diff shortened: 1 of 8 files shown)


beam-migrate/Database/Beam/Migrate.hs
```diff
@@ -11,11 +11,12 @@
...
 --
 -- A checked database is similar to the 'DatabaseSettings' type from
--- @beam-core@. Whereas a 'DatabaseSettings' object consists of a set of
+-- @beam-core@ except it comes with a set of /predicates/.
+-- Whereas a 'DatabaseSettings' object consists of a set of
 -- database entities, a checked database (represented by the
 -- 'CheckedDatabaseSettings' type) consists of a set of database entities along
 -- with a set of /predicates/ (represented by types implementing
--- 'DatabasePredicate').except it comes with a set of /predicates/.The
+-- 'DatabasePredicate').The
```
kmicklas, line 19: Missing a space between `.` and `The`.

PR #562, opened 2021-04-23
like_, similarTo_ and ilike_ support different types for haystack and…

alexbiehl:
… needle

Suppose you have a have strongly typed column URL and want to use like_, ilike_ or similarTo_ to query for a pattern of type Text. Currently that is not possible to express with Beam since e.g. creating an invalid URL from your pattern Text is not always an option - and that's what this patch fixes.

### DIFF
--- a/beam-core/Database/Beam/Query/Operator.hs
+++ b/beam-core/Database/Beam/Query/Operator.hs
@@ -59,16 +59,18 @@ infixr 3 &&., &&?.
 infixr 2 ||., ||?.
 
 -- | SQL @LIKE@ operator
-like_ :: ( BeamSqlBackendIsString be text
+like_ :: ( BeamSqlBackendIsString be left
+         , BeamSqlBackendIsString be right
          , BeamSqlBackend be )
-      => QGenExpr ctxt be s text -> QGenExpr ctxt be s text -> QGenExpr ctxt be s Bool
+      => QGenExpr ctxt be s left -> QGenExpr ctxt be s right -> QGenExpr ctxt be s Bool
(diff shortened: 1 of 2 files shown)


alexbiehl:
Gentle ping. It would be great if I could get some feedback wether this is wanted and the way to go.

kmicklas:
Thanks! This looks great, sorry for the delay.

I'll merge once the CI passes. (Currently GitHub is having an internal server error when I try to start it...)

alexbiehl:
Awesome, Thank you, Sir!

kmicklas:
I'm still seeing a 500 from GitHub but apparently my support request has been raised to the engineering team so we'll see if they can fix it. Otherwise if it still doesn't work after a couple more days I'll run the CI locally and merge.

alexbiehl:
I'll run the CI locally and merge.

How is it going?

kmicklas:
Sorry for the delay!

alexbiehl:
Thanks a lot! Would it be possible for you to cut a new release especially once https://github.com/haskell-beam/beam/pull/567 has been merged?

shlevy:
#567 merged now

kmicklas:
This turned out to be more more breaking than I imagined, since now all regular uses of these functions with static pattern strings will need a type annotation. I'm planning to make a new PR to add the abstracted versions as separate functions and keep the original types for the original functions. @alexbiehl do you have any thoughts on the naming? I'd like to avoid the default Beam convention which is append `'` because it's unclear but I can't think of anything not too verbose.

PR #563, opened 2021-04-29
Support nullary products

alexfmpe:
Needed this not for an empty table, but for an empty record which was then nested inside others to make a full table.


kmicklas:
Won't this allow fully empty tables to be declared? I'd like to see if there is a way to prevent that since it's not valid in SQL.

kmicklas:
I just confirmed that this let me declare an empty table and produce a `DatabaseSettings` for it.

@alexfmpe I'm going to close this PR for now since I don't want to make that possible, but feel free to reopen if you come up with a different approach. Maybe you can try adding another level of generic machinery so this isn't possible at the top level, but can be done inside nested tables?

cgibbard:
Note that PostgreSQL permits tables without columns and they're obviously a perfectly cromulent notion in relational algebra, but the SQL standards are more restrictive. Perhaps a backend-specific check could ensure that resulting tables conform to whatever the backend can handle?

```
postgres=# CREATE TABLE zero ();
CREATE TABLE
postgres=# \d zero
    Table "public.zero"
 Column | Type | Modifiers 
--------+------+-----------

postgres=# insert into zero default values;
INSERT 0 1
postgres=# SELECT * FROM zero;
--
(1 row)
```

Issue #411, opened 2019-11-06
Compatibility with recent version of resourcet with MonadUnliftIO


wavewave:
I found streaming interface of beam-postgres uses `MonadBaseControl` [link](http://hackage.haskell.org/package/beam-postgres-0.4.0.0/docs/Database-Beam-Postgres-Conduit.html#v:runSelect), while resourcet >= 1.2 moved to MonadUnliftIO. Any plan for updating the interface? Thanks.

shlevy:
Any thoughts on this? Would a patch be accepted?

LaurentRDC:
A patch would be accepted!

PR #567, opened 2021-06-06
Postgres.Conduit: Add streaming variants directly in a MonadResource

shlevy:
In addition to bringing compatability with the MonadBaseControl-free
conduit library, the new variants are much more composable (e.g. they
can be combined with servant's streaming capabilities).

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Conduit.hs
+++ b/beam-postgres/Database/Beam/Postgres/Conduit.hs
@@ -1,28 +1,51 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
 
 -- | More efficient query execution functions for @beam-postgres@. These
 -- functions use the @conduit@ package, to execute @beam-postgres@ statements in
 -- an arbitrary 'MonadIO'. These functions may be more efficient for streaming
 -- operations than 'MonadBeam'.
-module Database.Beam.Postgres.Conduit where
+module Database.Beam.Postgres.Conduit
+  ( streamingRunSelect
+  , runInsert
+  , streamingRunInsertReturning
+  , runUpdate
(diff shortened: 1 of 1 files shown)


shlevy:
Note that I haven't run the whole test suite or formatter (if any). Are there straightforward instructions for that? The nix shell isn't working on latest nixpkgs.

shlevy:
Fixes #411

beam-postgres/Database/Beam/Postgres/Conduit.hs
```diff
@@ -115,77 +177,119 @@ executeStatement conn x =
     syntax <- pgRenderSyntax conn x
     Pg.execute_ conn (Pg.Query syntax)
 
+
+-- | Runs any query that returns a set of values
+runQueryReturning'
+  :: ( C.MonadResource m, Fail.MonadFail m, FromBackendRow Postgres r )
+  => Pg.Connection -> PgSyntax
+  -> CONDUIT_TRANSFORMER () r m ()
+runQueryReturning' (conn@Pg.Connection {..}) x = do
```
kmicklas, line 186: Personally, I'd advise against `RecordWildCards` since it makes it impossible to know what's bound here without knowing the definition of `Pg.Connection`.

kmicklas (commented):
Can you think of any ideas for more descriptive names for the new variants than simply appending `'`? I'm not really a fan of the frequent use of that in Haskell in general, but especially here where the original variants are deprecated I feel like it is confusing.

shlevy:
@kmicklas Updated

kmicklas:
Any chance you can add some rudimentary tests? If it's too much of a burden I can merge and make an issue to add later.

shlevy:
@kmicklas I get test failures when I try to run, presumably because I don't have a local postgres instance. I think it would be too much to get to in the immediate future to get a test harness set up

PR #571, opened 2021-06-24
Add Postgresql specific EXTRACT fields

alexbiehl:


kmicklas (commented):
What's the value provided by the `IsPgExtractFieldSyntax` class? Since this is Postgres-specific and not abstracting over syntaxes I'm not sure what it would be useful for.

alexbiehl:
Good point, I removed the class.

PR #590, opened 2021-09-23
Add ExprAll and ExprAny monoids for Bool and SqlBool exprs

roberth:
Wrappers with `Monoid`  and `Semigroup` instances for boolean expression types, analogous with `Data.Monoid`'s `All` and `Any`.

Allows search criteria expressions to be composed monoidally, for example.

### DIFF
--- a/beam-core/test/Database/Beam/Test/SQL.hs
+++ b/beam-core/test/Database/Beam/Test/SQL.hs
@@ -47,6 +47,9 @@ tests = testGroup "SQL generation tests"
 
                   , noEmptyIns
 
+                  , exprAllTest
+                  , exprAnyTest
+
                   -- Regressions for github issues
                   , testGroup "Regression tests"
                     [ gh70OrderByInFirstJoinCausesIncorrectProjection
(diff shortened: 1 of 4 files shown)


beam-core/test/Database/Beam/Test/SQL.hs
```diff
@@ -1366,3 +1369,52 @@ gh70OrderByInFirstJoinCausesIncorrectProjection =
                    , selectLimit = Nothing , selectOffset = Nothing }
 
       actual @?= expected
+
+-- | Test ExprAll
+
```
kmicklas, line 1374: nit: here and below, haddocks should be adjacent to declaration

kmicklas (commented):
How about making `ExprAll` and `ExprAny` wrap a list, rather than an individual constructor. Then you'd provide a singleton function and a destructor which uses `mconcatMinimal`. (Maybe you can even hide the actual constructors.)

The current instances aren't actually monoidal and this is observable by the extra trues/falses that get inserted if you append `mempty`. In other words, it doesn't satisfy the identity laws.

beam-core/test/Database/Beam/Test/SQL.hs
roberth: It does look weird, doesn't it? I just followed the example of earlier definitions in this file, like https://github.com/haskell-beam/beam/pull/590/files#diff-f4e085f7289c80ab762596354429a86fc78598f1bd0ff87557d56527a392384fR1173

roberth:
> The current instances aren't actually monoidal

This depends on how we define equality on expressions. I assumed equivalent behavior in the database is sufficient. Using a list is a bit of a departure from `All` and `Any` in `base`, but I guess that's ok.

roberth:
Have to focus on other things and I don't feel particularly attached to this idea. Feel free to do whatever with this PR's code.

PR #597, opened 2021-10-15
Postgres.Conduit: throw SqlError when row is error

### LINKED ISSUE #596: `Postgres.Conduit` throws wrong type of exception, can't be caught properly
The conduit module uses

```haskell
Fail.fail ("Postgres error: " <> show errMsg)
```

which is like `throwIO . userError`.

This is a problem because postgresql-simple retry logic relies on `SqlError` exceptions when serialization needs to be retried.

I think all the ingredients are already there so I'll try to make a pr.

roberth:
Fixes #596 

Sadly I don't have a good reproducer for serialization errors, but I'm confident that this is the correct fix, based on the fact the error message had `Postgres error:` which only occurs on the line below. The three matches correspond to how postgresql-simple uses `throwResultError`.


roberth:
@kmicklas Is there anything else I can do?

kmicklas:
If you do find a way to write a test for this, another PR would be great, but I'll merge this as is.

PR #599, opened 2021-10-25
Add 'week' to date extraction

mankyKitty:
This change adds the `week_` field to the list of fields that can be extract from `HasSqlDate` compatible fields.

This will be a breaking change for anyone that manipulates or consumes the AST as there is a new field added to the `IsSql92ExtractFieldSyntax ` typeclass and a new constructor to `ExtractField`.


kmicklas:
Can you rebase to fix the CI?

PR #604, opened 2021-11-11
Allow aeson 2.0+ upgrade

dysinger:
Minimal changes to support Aeson 2.0+ (vulnerability upgrade)

### DIFF
--- a/beam-migrate/Database/Beam/Migrate/Serialization.hs
+++ b/beam-migrate/Database/Beam/Migrate/Serialization.hs
@@ -39,6 +39,9 @@ import           Control.Applicative
 import           Control.Monad
 
 import           Data.Aeson
+#if MIN_VERSION_aeson(2,0,0)
+import qualified Data.Aeson.Key as DAK
+#endif
 import           Data.Aeson.Types (Parser)
 import qualified Data.Dependent.Map as D
 import qualified Data.GADT.Compare as D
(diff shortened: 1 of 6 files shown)


beam-migrate/beam-migrate.cabal
```diff
@@ -62,7 +62,7 @@ library
   build-depends:       base                 >=4.9     && <5.0,
                        beam-core            >=0.9     && <0.10,
                        text                 >=1.2     && <1.3,
-                       aeson                >=0.11    && <1.6,
+                       aeson                >=0.11    && <1.6 || >=2.0.1.0 && <2.1,
```
kmicklas, line 65: Why is 2.0.0.0 not allowed? My understanding is the vulnerability applies to all previous versions.

beam-migrate/Database/Beam/Migrate/Serialization.hs
```diff
@@ -322,7 +325,11 @@ sql92Deserializers = mconcat
                    , beamDeserializer deserializeSql92Attributes ]
   where
     parseSub nm o key parse =
+#if MIN_VERSION_aeson(2,0,0)
+      withObject (unpack (nm <> "." <> key)) parse =<< o .: DAK.fromText key
```
kmicklas, line 329: nit: I prefer to avoid duplication in CPP selections as much as possible. So something like this:

```
#if MIN_VERSION_aeson(2,0,0)
    makeKey = DAK.fromText
#else
    makeKey = id
#endif
    parseSub nm o key parse =
      withObject (unpack (nm <> "." <> key)) parse =<< o .: makeKey key
```

beam-migrate/beam-migrate.cabal
dysinger: The press release said it was fixed in 2.0.1.0. I know. It's strange.

beam-migrate/Database/Beam/Migrate/Serialization.hs
dysinger: :+1:

beam-migrate/beam-migrate.cabal
kmicklas: But doesn't it apply to all 1.x versions as well? They use `HashMap` without even an option to switch to another implementation.

Also, I don't think it's the job of libraries to ban insecure downstream dependencies via version constraints. That's the job of the application developer / packaging system. Especially in this case, for many applications it is desirable to use `HashMap` anyway. If a user wants to use 2.0.0.0 I don't think Beam should be making that decision.

kmicklas:
Closing in favor of #617.

PR #626, opened 2022-01-27
`IN (SELECT ...)` syntax for beam

tathougies:
This is some basic functionality that's been missing for a while.

### DIFF
--- a/beam-core/beam-core.cabal
+++ b/beam-core/beam-core.cabal
@@ -2,7 +2,7 @@
 -- see http://haskell.org/cabal/users-guide/
 
 name:                beam-core
-version:             0.9.2.0
+version:             0.9.3.0
 synopsis:            Type-safe, feature-complete SQL query and manipulation interface for Haskell
 description:         Beam is a Haskell library for type-safe querying and manipulation of SQL databases.
                      Beam is modular and supports various backends. In order to use beam, you will need to use
--- a/beam-core/beam-core.cabal
+++ b/beam-core/beam-core.cabal
@@ -2,7 +2,7 @@
 -- see http://haskell.org/cabal/users-guide/
 
 name:                beam-core
-version:             0.9.3.0
+version:             0.10.0.0
 synopsis:            Type-safe, feature-complete SQL query and manipulation interface for Haskell
 description:         Beam is a Haskell library for type-safe querying and manipulation of SQL databases.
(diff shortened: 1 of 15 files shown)


beam-postgres/test/Database/Beam/Postgres/Test/Select.hs
```diff
@@ -99,6 +99,14 @@ testInRowValues getConn = testCase "IN with row values works" $
       return $ pair `in_` [pair, pair]
     assertEqual "result" [True] result
 
+testInSelect :: IO ByteString -> TestTree
```
kmicklas, line 102: It looks like this isn't in the test tree?

beam-core/beam-core.cabal
```diff
@@ -2,7 +2,7 @@
 -- see http://haskell.org/cabal/users-guide/
 
 name:                beam-core
-version:             0.9.2.0
+version:             0.9.3.0
```
kmicklas, line 5: I believe by the PVP this counts as a breaking change, even though it won't likely affect regular users. I wanted to get as much non-breaking stuff out of the backlog first before getting to breaking changes, but I think we're in a pretty good state now to start merging things for an eventual major release. Can you bump to 0.10?

beam-migrate/Database/Beam/Haskell/Syntax.hs
```diff
@@ -696,6 +696,7 @@ instance IsSql92ExpressionSyntax HsExpr where
   leE q a b = hsApp (hsVar "leE")   [hsMaybe q, a, b]
 
   inE a b = hsApp (hsVar "inE") [a, hsList b]
+  inSelectE _ _ = error "inSelectE"
```
kmicklas, line 699: I was going to ask why the `error` and then noticed there are already many more. :cry:

beam-sqlite/beam-sqlite.cabal
```diff
@@ -1,5 +1,5 @@
 name:                beam-sqlite
```
kmicklas, line 1: We now have select tests for SQLite, can you add one here too?

kmicklas (commented):
Thanks for doing this! I started working on this in 2019 but never got around to finishing it.

beam-core/beam-core.cabal
tathougies: You're right!

Issue #367, opened 2019-04-05
runSelectReturningOne should not return Nothing for more than one row


moll:
As asked on IRC, creating an issue for discussion.

`runSelectReturningOne` returns `Nothing` when more than one row matches the query.  https://github.com/tathougies/beam/issues/268 states this is by design.

I claim that returning `Nothing` when more than one row is returned is a risky choice. It's correct to state the row ordering may not be defined and non-determinism may result, but returning `Nothing` signifies there are no rows that match the given constraints. That's likely to lead code to attempt to create a row and then fail either with a database constraint failure, or to further exacerbate the duplicate row problem. Secondly, a lot of other ORMs and database layers do not behave like this. Be it an OO `.first` or a functional `getFirst`, they often tack on a `LIMIT 1` and get on with it. `runSelectReturningOne`, while not modifying the query to optimize with a `LIMIT 1`, shouldn't at least behave in an unexpecting manner. I, for one, leaned towards `runSelectReturningOne` to not have to do `listToMaybe`, and didn't think twice whether there are or aren't duplicate rows.

Cheers.

mulderr:
Semantics of `runSelectReturningOne` is that you expect exactly one row. Multiple rows in that case is equivalent to no rows - something unexpected, something very wrong with your assumptions. That is useful in it's own right and it did save me a couple times. I caught the error early which would not have happened had `runSelectReturningOne` silently done a `limit 1`.

What you describe I would rather call `runSelectReturningFirst` and you can easily get it combining `runSelectReturningList` and `listToMaybe` or so - as you observe yourself.

The reason there isn't something like `runSelectReturningFirst` in beam already is I guess exactly what you indicated above. It's not a very good primitive to begin with and could be "unstable" if not properly handled.

moll:
I alluded to above that the returned `Nothing` also has a meaning — a meaning of "nothing matched", not something out of the ordinary happened. If it's meant to say "something unexpected" happened, that should be encoded in the type system through an `Either MoreRowsError (Maybe row)`. I personally don't care to protect myself from more rows and will choose the get-one-and-ignore-the-rest choice. As for naming, I believe you, @mulderr, intuitively differentiate between `*First` and `*One`, but I suspect very few others share that intuition. :P As with web pages, we spend most of our time on other APIs, and all of those tend to behave as I described above — `.first`-likes gets the first row and ignores the rest.

3noch:
[...]

PR #630, opened 2022-02-03
Add runSelectReturningFirst (RFC)

thomasjm:
This is a proposal to address #367.

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL.hs
+++ b/beam-core/Database/Beam/Backend/SQL.hs
@@ -133,6 +133,11 @@ class (BeamBackend be, Monad m) =>
                  Nothing -> pure (Just x)
                  Just _ -> pure Nothing
 
+  -- | Run the given command and fetch the first result. The result is
+  --   'Nothing' if no results are returned.
+  runReturningFirst :: FromBackendRow be x => BeamSqlBackendSyntax be -> m (Maybe x)
+  runReturningFirst cmd = runReturningMany cmd id
+
   -- | Run the given command, collect all the results, and return them as a
   --   list. May be more convenient than 'runReturningMany', but reads the entire
   --   result set into memory.
--- a/beam-core/Database/Beam/Query.hs
+++ b/beam-core/Database/Beam/Query.hs
@@ -64,6 +64,7 @@ module Database.Beam.Query
     , select, selectWith, lookup_
     , runSelectReturningList
     , runSelectReturningOne
+    , runSelectReturningFirst
     , dumpSqlSelect
 
     -- ** @INSERT@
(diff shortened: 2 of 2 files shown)


kmicklas (commented):
This looks good and I like this approach.

Can you specialize `runSelectReturningFirst` for Postgres to use `AtOnce` instead of `CursorBatching`? See #520. This should then get a Postgres select test.

alexfmpe:
I'm slightly against the name - what does `First` really mean when there's no `order_` ? Is there some consistent default order or can we get different elements at different times? Does this change across backends?

My issue with `runSelectReturningOne` is that the naming doesn't point to the meaning of the query's result set and thus hides the whole "at least one" vs "exactly one" constraint on a success, so you often forget to consider whether `limit_ 1` should be used or not and get unexpected failures.
I've seen this happen like a dozen times across projects and people and think that issue will exist as long as `runSelectReturningOne` exists, which is why I'd prefer renaming it (via deprecation) to `runSelectReturningUnique` or some other in-your-face name that'd point to the constraint you're buying into, with `limit_ 1` as an obvious opt-out.

This seems like it'd lead to confusion between whether to pick `runSelectReturningFirst` and `runSelectReturningOne $ limit_ 1` - do they have different performance?

thomasjm:
> I'm slightly against the name - what does First really mean when there's no order_ ? Is there some consistent default order or can we get different elements at different times? Does this change across backends?

If the name causes the user to ask themselves questions like this, then I think it succeeds in its goal :). I'd love for  `runSelectReturningOne` to be removed or renamed, but that seems like a major breaking change. For the time being, I think adding `runSelectReturningFirst` is good because it draws attention to the distinction between the two. So hopefully people won't see the `Maybe` return type and blindly select `runSelectReturningOne` because it looks like what they want, but will instead read the docstrings on both functions and choose the right one.

kmicklas:
> I'm slightly against the name - what does First really mean when there's no order_ ? Is there some consistent default order or can we get different elements at different times? Does this change across backends?

To me it seems pretty clear that it just returns the first result, and ordering is up to you. It's a totally valid use case to just want *any* individual result and not care about the ordering.

The one possible confusion that makes sense to me is about whether this automatically adds a `LIMIT 1`. @thomasjm can you add a note in the haddock saying something like "This does not automatically apply a limit to the query, you likely want to do that if you expect many results."? However, if anything, I think users are more likely to have that confusion with `runSelectReturningOne`.

> why I'd prefer renaming it (via deprecation) to runSelectReturningUnique or some other in-your-face name that'd point to the constraint you're buying into, with limit_ 1 as an obvious opt-out.

I think this is a good idea, but we can keep it separate from this PR since it doesn't actually interact. If we do that renaming, `runSelectReturningFirst` still seems to make sense.

I'm planning to do a new major release cycle soon, so we can include it in that. What I'll likely do is rename the class method but add back `runSelectReturningOne` as a synonym function and mark it deprecated. This way user code should continue to work (with a warning) and only backends would need to update their instances.

thomasjm:
>  @thomasjm can you add a note in the haddock saying something like "This does not automatically apply a limit to the query, you likely want to do that if you expect many results."?

I was just looking at specializing `runSelectReturningFirst` for Postgres and wondered -- is there some reason that `runSelectReturningOne` doesn't automatically add a LIMIT? It seems it could do its job more efficiently with a LIMIT 2 applied.

Same reasoning applies to the new `runSelectReturningFirst`, maybe it would be correct and more efficient to just add LIMIT 1?

beam-core/Database/Beam/Backend/SQL.hs
```diff
@@ -133,6 +133,12 @@ class (BeamBackend be, Monad m) =>
                  Nothing -> pure (Just x)
                  Just _ -> pure Nothing
 
+  -- | Run the given command and fetch the first result. The result is
+  --   'Nothing' if no results are returned.
+  --   This is not guaranteed to automatically apply LIMIT 1 to the query.
```
tathougies, line 138: `LIMIT 1` is a postgresql-ism. I would use something neutral in the description.

thomasjm: Done!

PR #639, opened 2022-10-11
Update Tables.hs

tdimiduk:
I couldn't find `nothingRef` it looks like the actual value I want is `nothing_`. Assuming my interpretation is correct, changing the docs here would make it a little easier for the next person to figure this out.


kmicklas:
The sentence is still a bit confusing, but this is an improvement. Thanks!

Issue #646, opened 2022-11-29
Postgres lateral_ and guard_ generating incorrect SQL? (all results are aliased as `t0`)


maerten:
It seems that using Postgres `lateral_` and `guard_` results in incorrect SQL.

At the end of the generated query, I see a clause: 	
    
    WHERE ("t0"."res3") IS NOT DISTINCT FROM ("t0"."res5")
             ^--    (both referencing t0)    ---^

This is odd as the `guard_` clause references two different Entities with the `==.` operator.

Investigating this a bit further, it appears that all `CROSS JOIN LATERAL` end up with an alias `AS "t0"`.

kmicklas:
Can you post a minimal query that produces incorrect SQL?

maerten:
Here's a sample project which reproduces the issue: https://github.com/maerten/beam-issue-example

tathougies:
This is caused by `buildQuery` in `beam-core/Database/Beam/Query/SQL92.hs` assuming we can use the same table names in sub queries in joins, which is not a good assumption for lateral join. I think the proper approach is to thread the table prefix in `buildQuery` so that we can disambiguate the names. I will put up a PR

tathougies:
Just pushed PR #647. @maerten  would you mind verifying?

I did notice an error in your repository, you should use

```
deriving instance Show User
```

instead of

```
instance Show User
```

The former uses GHC's deriving mechanism to derive a proper implementation. The latter uses the default implementations in the `Show` class which are mutually recursive, thus leads to a stack overflow.

maerten:
@tathougies thanks for the fix, the fix works in the example project, and in another project too.

Expanding on the `lateral_` queries, what's the right place to add limit the results of a `lateral_` query?
I tried adding `limit_` as shown in the example at https://hackage.haskell.org/package/beam-postgres-0.5.2.1/docs/Database-Beam-Postgres-Full.html#g:3

However that results in a type error:
```
    • Couldn't match type ‘QBaseScope’
                     with ‘Database.Beam.Query.Internal.QNested QBaseScope’
```

maerten:
And one last question, it looks like `lateral_` always generates a `CROSS LATERAL JOIN`, even when using a `leftJoin_`.
This was unexpected to me, as the code for `lateral_` looks like it would create different type of join if there is an 'on' clause. But perhaps I'm misunderstanding things.. 

[...]

PR #647, opened 2022-12-02
Make sure lateral join names do not overlap

tathougies:
This is a fix for #646


tathougies:
@kmicklas I'm no longer familiar with how the docs are generated. How do we verify that the docs all build (and thus have some confidence that the changes I made don't have any other deleterious effects)?

PR #658, opened 2023-01-10
Fix the instance of SqlDeconstructMaybe for Beamable tables

cgibbard:
It was previously only considering a Maybe row to be Just in the case that none of the fields were Nothing. This changes the logic so that it instead checks that it's not the case that all the fields are Nothing (i.e. if anything is Just, the whole thing is).

This will be surely correct for any table which is not entirely Maybe / nullable. In the case when all the fields are Maybes, it still can't distinguish a record of all Nothings from a Nothing, but beam doesn't attempt to make that distinction elsewhere anyway.


cgibbard:
sorry, meant to make this against a fork

Issue #667, opened 2023-04-10
WHERE clauses are silently dropped if there is no FROM


kosmikus:
As the title says, a query such as `SELECT true WHERE false` is not rendered correctly by Beam. In general, `WHERE` clauses are dropped in queries where there is no `FROM`. The `WHERE` clause is quite meaningful though, as it determines whether the query has zero or one resulting rows.

This is a minimal example (the real context in which this appeared was significantly more complex):
```haskell
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE KindSignatures #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE OverloadedStrings #-}
{-# OPTIONS_GHC -Wall #-}
module BeamDebug where

import Data.Kind
import Database.Beam
import Database.Beam.Postgres
import Database.Beam.Schema.Tables

data EmptyDb (f :: Type -> Type) =
  EmptyDb

instance Database be EmptyDb where
  zipTables _ _ _ _ = pure EmptyDb

simple :: Q Postgres EmptyDb s (QExpr Postgres s Bool)
simple = do
  guard_ (val_ False)
  pure (val_ True)

test :: IO ()
test = do
  conn <- connectPostgreSQL "dbname=beamtestdb"
  pgTraceStmtIO conn (select simple)

-- output:
-- SELECT true AS "res0"
--
-- expected:
-- SELECT true AS "res0" WHERE false
```

I've been looking at the source code, and the problem seems to be in `Database.Beam.Query.SQL92`:
```haskell
buildSql92Query' :: forall be db s a
                  . ( BeamSqlBackend be, Projectible be a)
                 => Bool {-^ Whether this backend supports arbitrary nested UNION, INTERSECT, EXCEPT -}
                 -> T.Text {-^ Table prefix -}
                 -> Q be db s a
                 -> BeamSqlBackendSelectSyntax be
buildSql92Query' arbitrarilyNestedCombinations baseTblPfx (Q q) =
    buildSelect baseTblPfx (buildQuery baseTblPfx (fromF q))
  where
    be :: Proxy be
    be = Proxy

    buildQuery :: forall s x
                . Projectible be x
               => T.Text
               -> Free (QF be db s) x
               -> SelectBuilder be db x
    buildQuery _ (Pure x) = SelectBuilderQ x emptyQb
[...]

PR #695, opened 2023-11-07
Fix building queries with WHERE in the absence of FROM.

kosmikus:
A WHERE clause in a query is meaningful even if there are no input tables. It makes the difference between there being 0 or 1 results. We therefore should not drop guards when building queries.

See #667.


kmicklas (commented):
Can you add a test?

LaurentRDC:
I will add a test case after merge

LaurentRDC:
This PR has been released in [beam-core-0.10.3.0](https://hackage.haskell.org/package/beam-core-0.10.3.0)

PR #696, opened 2024-01-04
Improved error for ColumnTypeMismatch

thomasjm:
I was getting `ColumnTypeMismatch` errors and finding them to be pretty opaque if you don't know what column is causing the problem.

It turns out the [Postgres error types](https://hackage.haskell.org/package/postgresql-simple-0.7.0.0/docs/Database-PostgreSQL-Simple.html#t:ResultError) have an `errSQLField` which is useful to expose in the error message.


LaurentRDC:
Looks very sensible, thanks!

PR #707, opened 2024-06-03
Changed documentation example of withDbModification function.

tusharad:
## Issue
In the documentation example of the `withDbModification` function, the `modifyTable` function was used, but it has been deprecated. Additionally, the closing parenthesis was missing.

## Solution

I modified the example to use `setEntityName` and `modifyTableFields` as recommended.

Do let me know, if any changes are required. Thanks!

@kmicklas


LaurentRDC:
Awesome, thank you!

Issue #662, opened 2023-02-27
Requesting `subselectWith_` combinator


enobayram:
As far as I can tell, currently the only way to write recursive CTEs with `beam` is to use the `selectWith` combinator, which can only happen at the "top-level". However, (at least) Postgres supports recursive CTEs inside subqueries as well:

``` SQL
SELECT * 
FROM (
  WITH RECURSIVE my_cte AS (
    SELECT 1 
    UNION ALL 
    SELECT 1 WHERE FALSE
  ) 
  SELECT * 
  FROM my_cte
) AS t;

 ?column? 
----------
        1
(1 row)
```

Would it be possible to write a `subselectWith_` primitive, analogous to `subquery_` and `selectWith_` that allows CTEs inside subselects like this (without even exposing the `With` wrapper outside the subselect)?

enobayram:
Seems relevant to mention here that I've given up on type safety and managed to work around this using a combination of `pgUnnest` and `customExpr_` in [this commit](https://github.com/kadena-io/chainweb-data/pull/66/commits/26c7d8227d2683ae96dd4401b3545c1bef2bac89). Looking back at my workaround, I don't think even a `subselectWith_` would've solved my problem in combination with `lateral_`, because the `QExpr be (QNested s)` I get inside the `lateral_` query wouldn't be usable inside the `subselectWith_`, which would expect a `Q be db (QNested (QNested s))`, so I'd actually need something like a `lateralWith_`. A tangent for this GitHub issue, but If I'm not mistaken, I should get a special `QExpr be s` inside the `lateral_` that I can lift into as many layers of `QNested` as I want in order to fully capture the expressiveness of Postgres' `LATERAL` joins.

tathougies:
I know it's very late, but one part of your requested functionality is in the issue mentioned above. As for the renesting within `lateral_`, let me think about that one. It's a bit hard to express using our scoping mechanism.

LaurentRDC:
Fixed in #720

LaurentRDC:
I mistakenly thought that #720 fixed this issue, but it only addresses part of it. My apologies

PR #720, opened 2024-10-18
Allow inner subqueries with CTEs in postgres

tathougies:
This is a response to #662 which allows postgres (which supports inner CTEs) to embed `With` monads into regular `Q` objects, instead of just `SqlSelect` (which is what `selectWith` does). The `selectWith` behavior is correct since that's the standard specified constraint, but `pgSelectWith` is more flexible.

Docs are included (working on building them again)


beam-postgres/Database/Beam/Postgres/Connection.hs
```diff
@@ -155,13 +155,13 @@ runPgRowReader conn rowIdx res fields (FromBackendRowM readRow) =
                             Pg.ConversionFailed { Pg.errSQLType = sql
                                                 , Pg.errHaskellType = hs
                                                 , Pg.errMessage = msg
-                                                , Pg.errSQLField = field } ->
```
tathougies, line 158: Getting rid of GHC warnings here.

beam-postgres/Database/Beam/Postgres/Full.hs
```diff
@@ -274,6 +280,41 @@ lateral_ using mkSubquery = do
...
+pgSelectWith (CTE.With mkQ) =
+    let (q, (recursiveness, ctes)) = evalState (runWriterT mkQ) 0
+        fromSyntax tblPfx =
+            case recursiveness of
+              CTE.Nonrecursive -> withSyntax ctes (buildSqlQuery tblPfx q)
+              CTE.Recursive -> withRecursiveSyntax ctes (buildSqlQuery tblPfx q)
+    in Q (liftF (QAll (\tblPfx tName ->
+                           let (_, names) = mkFieldNames @Postgres @res (qualifiedField tName)
+                           in fromTable (PgTableSourceSyntax $
+                                         mconcat [ emit "(", fromPgSelect (fromSyntax tblPfx), emit ")" ])
```
tathougies, line 310: Need the `( )` to surround the `WITH` clause here.

docs/user-guide/backends/beam-postgres.md
```diff
@@ -274,3 +274,57 @@ runInsert $
...
+As an example using our Chinook schema, suppose we had an error with all orders in the month of
+September 2024, and needed to send out employees to customer homes to correct the issue. We want to
+find, for each order, an employee who lives in the same city as the customer, but we only want the
+highest ranking employee for each customer.
+
+First, we order the employees by org structure so that managers appear first, followed by direct reports. We use a recursive query for this, and then join it against the orders.
+
+!beam-query
+```haskell
+!example chinook only:Postgres
```
tathougies, line 307: This actually works!

LaurentRDC:
Tangential comment: what's the process for generating documentation now?
Most importantly, looks like the documentation is technically also part of tests, so it would be important to add this to the CI pipeline.

tathougies:
I have a nix expression I use to build documentation. I know ken had some issues with using Nix, but my own workflow is nix heavy. I could try doing it with Docker and seeing if I get it to work. It unfortunately needs a postgresql server as well.

LaurentRDC:
Thanks! I'll try to add a test case in the test suite as well after merging

Issue #725, opened 2024-11-18
remove "`beam-migrate` tool" from docs


peterbecich:
Following https://github.com/haskell-beam/beam/pull/684

it is still here https://haskell-beam.github.io/beam/schema-guide/tool/

![image](https://github.com/user-attachments/assets/5128a3a3-9caa-4e15-8183-1c43346f0d04)

The markdown file has already been deleted: https://github.com/haskell-beam/beam/edit/master/docs/schema-guide/tool.md
Does something need to be regenerated?

LaurentRDC:
Something does need to be regenerated. This is not currently automated using GitHub actions, but it should

LaurentRDC:
Closed via #727

PR #726, opened 2024-11-18
document `beam-migrate` as a dependency rather than an executable

peterbecich:
`beam-migrate` CLI tool has been deleted: https://github.com/haskell-beam/beam/pull/684


LaurentRDC:
Thanks for cleaning that up.

Looking at the documentation online, I also see a page about the `beam-migrate` tool (see below). Could you also remove all references to this page in the table of contents?

image

LaurentRDC:
Ahh nevermind, this is addressed in #725

PR #733, opened 2024-12-23
demonstrate `alterTable` and `addColumn` in example migration

peterbecich:
This PR improves the `V0002` example migration to include a demo of `alterTable` and `addColumn`.

The example already demonstrated the `preserve` function but not these two other functions.

Now the `V0002` migration additionally adds a `salary` field to `Staff`.

Here is the addition to the migration, rendered to SQL:
```
SQL command type: PgCommandTypeDdl
SQL script: 
ALTER TABLE "staff" ADD COLUMN "salary" SMALLINT DEFAULT 100 NOT NULL
```
Printing this depends on https://github.com/haskell-beam/beam/pull/732


LaurentRDC:
Good idea, thanks!

PR #736, opened 2024-12-31
provide `Generic` instances for `...SQL.Types`

peterbecich:
Helps with QuickCheck on datatypes built on top of these built-in types


LaurentRDC:
Thanks!

PR #737, opened 2025-01-02
replace `nix-shell` with Nix Flake

peterbecich:
This `flake.nix` provides the same resources as the deleted Nix files with less code to maintain, plus other tools helpfully provided by https://github.com/srid/haskell-flake.

This commit does not fix/update the docs generation via Nix; this remains TODO.

The `flake.nix` will build bundles for each Beam library, just as the old Nix configuration did:
```
$ nix flake show
...
    └───x86_64-linux
        ├───all: package 'all'
        ├───beam-core: package 'beam-core-0.10.3.1'
        ├───beam-migrate: package 'beam-migrate-0.5.3.1'
        ├───beam-postgres: package 'beam-postgres-0.5.4.1'
        ├───beam-sqlite: package 'beam-sqlite-0.5.3.1'
        ├───default: package 'all'
        └───pagila: package 'pagila-0.1.0.0'
```
`nix build` or `nix build .#beam-...` will build these.

Beam has no `executables` so there is nothing to demonstrate by `nix run .#...`

I need more time to understand the docs generation, which apparently requires a Beam executable. This may have been the `beam-migrate` executable, which no longer exists. Please review this; I will try to return the the docs as a follow-up item.


.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
peterbecich, line 1: Tracking the `.envrc` in version control is debated; personally I am in favor of it

.github/workflows/nix-flake.yaml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - name: "Check `nix develop` shell"
+        run: nix develop --check
+      - name: "Check `nix develop` shell can run command"
+        run: nix develop --command "echo"
+      - run: nix build
+      - run: nix flake check
```
peterbecich, line 17: check the `flake.nix` and its outputs are not broken

docs/default.nix
```diff
@@ -4,8 +4,6 @@
 with nixpkgs;
```
peterbecich, line 4: this file is made dead code by this PR. I want to follow-up on this later.

LaurentRDC:
I don't use Nix so I will leave the review of this PR to someone else

peterbecich:
@kmicklas please review. I will fix `docs/default.nix` later.

.envrc
kmicklas: Personally I would prefer not to commit to this, since a lot more people have direnv than Nix, and Nix isn't required to easily develop on this repo.

.github/workflows/nix-flake.yaml
```diff
@@ -0,0 +1,24 @@
...
+    name: Nix on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - name: "Check `nix develop` shell"
+        run: nix develop --check
+      - name: "Check `nix develop` shell can run command"
+        run: nix develop --command "echo"
```
kmicklas, line 22:
```suggestion
        run: nix develop --command "true"
```
nit: this seems slightly more clear/idiomatic to me?

flake.nix
```diff
@@ -0,0 +1,63 @@
...
+
+           tools = hp: {
+             inherit (pkgs)
+               postgresql
+               sqlite-interactive
+               poetry
+               curl
+               pv # http://www.ivarch.com/programs/pv.shtml
+             ;
+
```
kmicklas, line 42:
```suggestion
```
nit: remove empty line

flake.nix
```diff
@@ -0,0 +1,63 @@
...
+            testcontainers.check = false;
+            beam-postgres.check = false;
+          };
+
+          devShell = {
+           enable = true;
+
+           hlsCheck.enable = false;
+
+           tools = hp: {
```
kmicklas, line 34: Is `hp` `haskellPackages`? If so I would prefer it to be more explicit for readability here.

kmicklas (approved):
Thanks for getting the CI to work with this!

Personally I am a flakes conscientious objector so unfortunately I cannot really review if the flake setup looks good. However from general Nix usage this all looks very nice.

PR #739, opened 2025-01-03
Remove direct dependence of `beam-sqlite` on platform-specific packages

### LINKED ISSUE #738: Wasm support
Hi! We're looking for a Selda replacement, and Beam looks promising.

We'd like to use `beam-sqlite` with GHC's Wasm backend, but it currently fails to build due to missing `getProcessID` support (or equivalent):

https://github.com/haskell-beam/beam/blob/df0080ec90b3fe41e89218198c3fbc3822820d89/beam-sqlite/Database/Beam/Sqlite/Connection.hs#L393

It appears Beam uses the process ID only as a unique identifier for the purposes of tracking a particular savepoint in `runInsertReturningList`. Could `getProcessID` potentially be replaced in Wasm by a pseudo-random number, instead?

LaurentRDC:
This PR removes `beam-sqlite`'s direct dependence on either the `unix` or `windows` package, which should allow to build `beam-core` and `beam-sqlite` on a wider variety of platforms.

Fixes #738  

Before merging, I want to at least manually check that `beam-sqlite` can be compiled using the WASM backend

### DIFF
--- a/beam-sqlite/ChangeLog.md
+++ b/beam-sqlite/ChangeLog.md
@@ -1,23 +1,30 @@
+# 0.5.4.0
+
+## Added features
+
+ * Removed the reliance on either the `unix` or `windows` package, which should enable (#738)
+   `beam-sqlite` to be buildable on a wider variety of platforms.
+
 # 0.5.3.1
 
-# Added features
+## Added features
 
  * Replaced use of deprecated functions.
 
 # 0.5.3.0
 
-# Added features
+## Added features
 
  * Loosen some version bounds
  * `HasSqlEqualityCheck` instance for `Day`
 
 # 0.5.2.0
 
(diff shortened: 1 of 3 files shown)


dhess:
Thanks for the remarkably quick turnaround on this request!

We haven't yet properly integrated `beam-sqlite` into our project, but I quickly added it to our project's deps so that it gets built, and I can confirm that this branch builds with the Wasm backend, using the latest version of these tools: https://gitlab.haskell.org/haskell-wasm/ghc-wasm-meta

LaurentRDC:
Thanks for checking! [I released a new version of `beam-sqlite` on Hackage](https://hackage.haskell.org/package/beam-sqlite)

dhess:
Amazing, thank you!

Issue #692, opened 2023-08-27
Postgres column doesn't roundtrip for type `Maybe (Vector Text)`


roberth:
I have a table with

```haskell
    labels :: Columnar f (Maybe (Vector Text))
```

Inserts work fine. `Nothing` and `Just []` are distinct in the database, as expected.
However, queries return `Just []` when they should `Nothing`.

This might be a problem with the `FromBackendRow` instance for `Maybe` and/or `Vector`.

Found on beam version 6b3e4c91ebabafc79878a35648d7f78d378ce445

LaurentRDC:
The issue is because of this line: https://github.com/haskell-beam/beam/blob/15af356e3402392d990ca1493ac9ce001da054cf/beam-postgres/Database/Beam/Postgres/Types.hs#L162

There is a special switch when parsing `Vector`, but in this instance, we should never expect a null. Pull request incoming

LaurentRDC:
The fix for this issue was released in [beam-postgres 0.5.4.2](https://hackage.haskell.org/package/beam-postgres-0.5.4.2)

PR #741, opened 2025-01-05
Support Postgres array migrations

### LINKED ISSUE #354: `error` call in array migration
I am attempting to do an initial schema migration (generate database from empty schema) and it is complaining that it cannot do array migrations, It appears that the offending `error` call is here:

```haskell
pgUnboundedArrayType :: PgDataTypeSyntax -> PgDataTypeSyntax
pgUnboundedArrayType (PgDataTypeSyntax _ syntax serialized) =
    PgDataTypeSyntax (error "Can't do array migrations yet")
                     (syntax <> emit "[]")
                     (pgDataTypeJSON (object [ "unbounded-array" .= fromBeamSerializedDataType serialized ]))
```

Is there any reason I can't just replace `error` with an actual description or oid and PR that in, or is that change substantially more involved than it first appears? At the moment I only need it to do the trivial migrations for me, so this seems like it would be OK.

LaurentRDC:
This pull request adds the ability to migrate Postgres' Array types.

Note that the mechanism by which this is done is still unsafe; if trying to migrate to an array type whose elements are unsupported (i.e. array OIDs that [don't exist here](https://hackage.haskell.org/package/postgresql-simple-0.7.0.0/docs/Database-PostgreSQL-Simple-TypeInfo-Static.html)), then the migration will result in a runtime error rather than a compile time error.

Breaking changes would have to be made to get compile-time safety of array migrations. 

Fixes #354

### DIFF
--- a/beam-postgres/test/Database/Beam/Postgres/Test/Marshal.hs
+++ b/beam-postgres/test/Database/Beam/Postgres/Test/Marshal.hs
@@ -51,6 +51,10 @@ boxGen = do PgPoint x1 y1 <- pointGen
             pure (PgBox (PgPoint (min x1 x2) (min y1 y2))
                         (PgPoint (max x1 x2) (max y1 y2)))
 
+arrayGen :: Hedgehog.Gen a -> Hedgehog.Gen (Vector.Vector a)
+arrayGen = fmap Vector.fromList 
+         . Gen.list (Range.linear 0 5) -- small arrays == quick tests
+
 boxCmp :: PgBox -> PgBox -> Bool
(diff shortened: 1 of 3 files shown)


beam-postgres/test/Database/Beam/Postgres/Test/Marshal.hs
```diff
@@ -164,70 +182,3 @@ marshalTest' cmp gen postgresConn =
       v' `cmp` a
 
     assertBool "Hedgehog test failed" passes
-
```
LaurentRDC, line 167: This test from #692 become redundant now

PR #750, opened 2025-03-06
Updates to migration backends

tathougies:
After several years, I've finally started working on a tool for beam migrations. I'll start a new repository for my tool, but there are some pieces missing from upstream beam that would be necessary for any migrations tool.

The key change here is adding a function to each backend to execute arbitrary SQL. This is necessary for any tool that allows running arbitrary migration scripts stored as backend-specific SQL.

The other major change is proper support for backend-agnostic transactions and proper connection handling.

All these changes are necessary for any migrations frontend, not just my tool.

### DIFF
--- a/beam-migrate/Database/Beam/Migrate/Backend.hs
+++ b/beam-migrate/Database/Beam/Migrate/Backend.hs
@@ -26,7 +26,7 @@
 --
 -- For an example migrate backend, see "Database.Beam.Sqlite.Migrate"
 module Database.Beam.Migrate.Backend
-  ( BeamMigrationBackend(..)
+  ( BeamMigrationBackend(..), BeamMigrateConnection(..)
   , DdlError
 
   -- * Haskell predicate conversion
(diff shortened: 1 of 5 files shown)


beam-migrate/Database/Beam/Migrate/Backend.hs
```diff
@@ -88,9 +88,16 @@ data BeamMigrationBackend be m where
     , backendFileExtension :: String
     , backendConvertToHaskell :: HaskellPredicateConverter
     , backendActionProvider :: ActionProvider be
-    , backendTransact :: forall a. String -> m a -> IO (Either DdlError a)
+    , backendRunSqlScript :: Text -> m ()
```
LaurentRDC, line 91: Writing SQL manually can lead to errors.
How about having `backendRunSqlScript :: Text -> m (Either (SqlException be) ())` or something?

For example, `postgresql-simple`'s `execute` can [throw exceptions](https://hackage.haskell.org/package/postgresql-simple-0.7.0.0/docs/Database-PostgreSQL-Simple.html#v:execute)

LaurentRDC:
That is a very good idea. At work we use `dbmate`, but it would be cool to be able to transition slowly from hand-written SQL migration scripts to the ORM.

beam-migrate/Database/Beam/Migrate/Backend.hs
tathougies: Regarding exceptions: all these monad actions run in `m` (which is an existential type here) which can only be executed by `backendRun` in `BeamMigrateConnection`, which returns an `Either` type. I agree that `DdlError` (thin wrapper over a string) is not the right type, but the main concern that exceptions could be thrown is handled by the fact that all these actions need to be run by this potentially-failable runner.

tathougies: And then for SQL scripts... I've tried a variety of ways of doing this, and for DDL at least, I think it's necessary to allow fallbacks to SQL scripts. There's too many databases, and options to realistically support this as a DSL. We already cover the basic DDL languages via beam-migrate.

Issue #326, opened 2018-11-08
VALUES produces invalid SQLite query


bgamari:
This example:

```haskell
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE TypeSynonymInstances #-}
{-# LANGUAGE StandaloneDeriving #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE UndecidableInstances #-}

module Main where

import GHC.Generics
import Data.Text (Text)
import Database.Beam
import Database.Beam.Sqlite
import Database.Beam.Backend.SQL
import Database.Beam.Backend.Types
import Database.SQLite.Simple as SQLite
import Database.SQLite.Simple.FromField

data TestDb f
    = TestDb
    { _tdbUsers          :: f (TableEntity UserT)
    }
    deriving (Generic)

instance Database be TestDb

testDb :: DatabaseSettings be TestDb
testDb = defaultDbSettings

type Username = Text

mkUsername = id

--newtype Username = Username Text
--                    deriving (Show, Eq, Ord)
--
--deriving instance HasSqlValueSyntax be Text => HasSqlValueSyntax be Username
--deriving instance FromField Username
--instance FromBackendRow Sqlite Username
--mkUsername :: Text -> Username
--mkUsername = Username

data UserT f
    = User
    { _userId   :: Columnar f Int
    , _userName :: Columnar f Username
    }
    deriving (Generic)

type User = UserT Identity
type UserId = PrimaryKey UserT Identity
deriving instance Show (PrimaryKey UserT Identity)
deriving instance Show User

instance Table UserT where
    data PrimaryKey UserT f = UserId (Columnar f Int) deriving Generic
    primaryKey = UserId . _userId

instance Beamable UserT
instance Beamable (PrimaryKey UserT)

missingUsers :: [Username]
             -> Q Sqlite TestDb s (QGenExpr QValueContext Sqlite s Username)
missingUsers users =
    let userNames :: Q Sqlite TestDb s (QGenExpr QValueContext Sqlite s Username)
        userNames = fmap _userName (all_ $ _tdbUsers testDb)

        userNames' :: Q Sqlite TestDb s (QGenExpr w Sqlite s Username)
[...]

PR #755, opened 2025-03-11
Rework VALUES clause and fix a bug on SQLite

tathougies:
This fixes an old bug #326 by factoring out all SQLite VALUEs into common table expressions, which seems to be the approach the SQLite manual takes to naming VALUE columns.

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL/AST.hs
+++ b/beam-core/Database/Beam/Backend/SQL/AST.hs
@@ -479,7 +479,7 @@ instance IsSql92TableNameSyntax TableName where
 data TableSource
   = TableNamed TableName
   | TableFromSubSelect Select
-  | TableFromValues [ [ Expression ] ]
+  | TableFromValues Int [ [ Expression ] ]
   deriving (Show, Eq)
 
 instance IsSql92TableSourceSyntax TableSource where
--- a/beam-core/Database/Beam/Query/Combinators.hs
+++ b/beam-core/Database/Beam/Query/Combinators.hs
@@ -112,11 +112,17 @@ values_ :: forall be db s a
            , BeamSqlBackend be )
         => [ a ] -> Q be db s a
 values_ rows =
-    Q $ liftF (QAll (\tblPfx -> fromTable (tableFromValues (map (\row -> project (Proxy @be) row tblPfx) rows)) . Just . (,Just fieldNames))
-                    (\tblNm' -> fst $ mkFieldNames (qualifiedField tblNm'))
+    Q $ liftF (QAll (\tblPfx -> fromTable (tableFromValues colCount (map (\row -> project (Proxy @be) row tblPfx) rows)) . Just . (,colAliases))
+                    (\tblNm' -> if useAliases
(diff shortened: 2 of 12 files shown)


beam-core/Database/Beam/Query/Combinators.hs
```diff
@@ -112,11 +112,17 @@ values_ :: forall be db s a
...
-    Q $ liftF (QAll (\tblPfx -> fromTable (tableFromValues (map (\row -> project (Proxy @be) row tblPfx) rows)) . Just . (,Just fieldNames))
-                    (\tblNm' -> fst $ mkFieldNames (qualifiedField tblNm'))
+    Q $ liftF (QAll (\tblPfx -> fromTable (tableFromValues colCount (map (\row -> project (Proxy @be) row tblPfx) rows)) . Just . (,colAliases))
+                    (\tblNm' -> if useAliases
+                                then fst $ mkFieldNames (qualifiedField tblNm')
+                                else fst $ mkDefaultFieldNames (qualifiedField tblNm'))
                     (\_ -> Nothing) snd)
     where
+      useAliases = knownBool @(BeamSqlBackendSupportsColumnAliases be)
+      colAliases | useAliases = Just fieldNames
```
LaurentRDC, line 122: Is there a reason to use type-level bools for this? Seems like a typeclass method using either `mkFieldNames` or `mkDefaultFieldNames` might be more straightforward

beam-core/Database/Beam/Backend/SQL/AST.hs
```diff
@@ -479,7 +479,7 @@ instance IsSql92TableNameSyntax TableName where
 data TableSource
   = TableNamed TableName
   | TableFromSubSelect Select
-  | TableFromValues [ [ Expression ] ]
+  | TableFromValues Int [ [ Expression ] ]
```
LaurentRDC, line 482: Would love to see a little comment here about what this `Int` represents

beam-sqlite/test/Database/Beam/Sqlite/Test/Select.hs
```diff
@@ -14,9 +14,10 @@ import Database.Beam.Sqlite.Test
 
 tests :: TestTree
 tests = testGroup "Selection tests"
-  [ expectFail testExceptValues
+  [ testExceptValues
```
LaurentRDC, line 17: Awesome!

LaurentRDC (approved):
This is great!

beam-core/Database/Beam/Query/Combinators.hs
tathougies: Good point.. yeah that would be easier

beam-core/Database/Beam/Backend/SQL/AST.hs
tathougies: This is the number of columns in the result (in the case that the expressions list is empty, we wouldn't be able to infer that from the values).

LaurentRDC: I meant as documentation, just like you did in `tableFromValues`

Issue #664, opened 2023-03-11
Lens example in tutorial does not work


tdimiduk:
The lens generation example in the docs appears to be malformed. `UserId` seems to be used as a data constructor in the pattern match, but it is not one:

```
Prelude> :load Tutorial.hs 
[1 of 1] Compiling Main             ( Tutorial.hs, interpreted )

Tutorial.hs:39:10: error:
    Not in scope: data constructor ‘UserId’
    Perhaps you meant ‘User’ (line 8)
   |
39 |         (UserId (LensFor addressForUserId)) =
```

Where I have made a file called `Tutorial.hs` that pulls in the relevant lines from tutorials 1 and 2

`Tutorial.hs`
```
{-#  LANGUAGE ImpredicativeTypes #-}
{-# LANGUAGE StandaloneDeriving #-}
import Database.Beam
import GHC.Int
import Data.Text (Text)

data UserT f
    = User
    { _userEmail     :: Columnar f Text
    , _userFirstName :: Columnar f Text
    , _userLastName  :: Columnar f Text
    , _userPassword  :: Columnar f Text }
    deriving Generic
type User = UserT Identity
type UserId = PrimaryKey UserT Identity

data AddressT f = Address
                { _addressId    :: C f Int32
                , _addressLine1 :: C f Text
                , _addressLine2 :: C f (Maybe Text)
                , _addressCity  :: C f Text
                , _addressState :: C f Text
                , _addressZip   :: C f Text
                , _addressForUser :: PrimaryKey UserT f }
                  deriving (Generic, Beamable)

type Address = AddressT Identity
deriving instance Show (PrimaryKey UserT Identity)
deriving instance Show Address

instance Table AddressT where
    data PrimaryKey AddressT f = AddressId (Columnar f Int32) deriving (Generic, Beamable)
    primaryKey = AddressId . _addressId
type AddressId = PrimaryKey AddressT Identity -- For convenience

Address (LensFor addressId)    (LensFor addressLine1)
        (LensFor addressLine2) (LensFor addressCity)
        (LensFor addressState) (LensFor addressZip)
        (UserId (LensFor addressForUserId)) =
        tableLenses

User (LensFor userEmail)    (LensFor userFirstName)
     (LensFor userLastName) (LensFor userPassword) =
     tableLenses
```

augusto:
[...]

Issue #659, opened 2023-01-12
LenseFor


yaitskov:

yaitskov:
I figured out how to define lenses manually as a workaround:

```
fooName ::
  forall (f1 :: Type -> Type) (f2 :: Type -> Type).
  Functor f2 =>
  (Columnar f1 Text -> f2 (Columnar f1 Text)) -> FooT f1 -> f2 (FooT f1)
fooName fa u = (\x -> u { _name = x } ) <$> fa (_name u)

```

peterbecich:
I have also seen this here which uses GHC 9.2: https://github.com/haskell-beam/beam/pull/650
The errors can be seen with `nix-build release.nix`

-------------------

@gelisam helpfully directed me to look more closely at the `ImpredicativeTypes` extension.

The snippets rely on it: 
https://github.com/haskell-beam/beam/blob/94461937c5ca8b89f1ff1a60bde5bffe207315b4/docs/tutorials/tutorial2.md?plain=1#L237-L240

This says `ImpredicativeTypes` was unreliable before GHC 9.2: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/impredicative_types.html 
Now perhaps the extension has changed?

berberman:
I hit the same issue at 9.2.7. Wiring the type signature by hand still (with `NoMonomorphismRestriction` enabled) doesn't work. I tried `fooName :: forall f. Functor f => Lens' (FooT f) (C f Text)`, though it compiles, but at the use site there is an error `No instance for (Functor ...)`. I also tried expanding `Lens'` further: `fooName :: forall f1 f2. Functor f2 => (C f1 Text -> f2 (C f1 Text)) -> FooT f1 -> f2 (FooT f1)`, but this doesn't compile due to `Could not deduce (Functor f1) arising from GHC Bug #20076`.

arguri:
@berberman and @yaitskov 

Have you tried 

```haskell

Module Foo where 

import Database.Beam 
import Lens.Micro 

data FooT f = Foo  { _fooInt :: C f Int }  deriving (Generic, Beamable)
type Foo = FooT Identity 

fooName :: Lens Foo Foo Int Int 
Foo  (LensFor fooName)  = tableLenses
```

yaitskov:
@arguri, it only works for data with 1 field (i.e. newtype)


```
• Could not deduce (Functor f0) arising from GHC Bug #20076
  from the context: Functor f
[...]

Issue #612, opened 2021-12-21
Publish docs from CI


kmicklas:
We have a Nix derivation to build the docs. Pushing to master (or maybe a release branch) should upload to GitHub pages.

A prerequisite for this is getting all the examples to compile again.

tathougies:
Which examples don't work? I can go through them.

kmicklas:
I fixed the failures in #613 so all that's left is to automate publishing.

PR #757, opened 2025-04-02
Update lens generation documentation

neuroevolutus:
This PR updates the documentation regarding lens generation to reflect the extra pattern matches that appear to be necessary with newer versions of GHC (as described in [#659](https://github.com/haskell-beam/beam/issues/659) and [#664](https://github.com/haskell-beam/beam/issues/664)).

### DIFF
--- a/docs/tutorials/tutorial2.md
+++ b/docs/tutorials/tutorial2.md
@@ -223,15 +223,18 @@ We can bring these lenses into scope globally via a global pattern match against
 -- Add the following to the top of the file, for GHC >8.2
 {-#  LANGUAGE ImpredicativeTypes #-}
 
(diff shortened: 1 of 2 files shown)


LaurentRDC:
Thanks!

Note that cocumentation isn't currently automatically published (#612), so your changes might not appear in the short term

neuroevolutus:
Just wanted to express my gratitude for your awesome work on maintaining this library! I've learned a lot by just using it.

LaurentRDC:
Thanks Abel. I want to point out that [many people contribute here](https://github.com/haskell-beam/beam/graphs/contributors). The original creator is still around as well!

Don't hesitate to raise more issues / submit more pull requests as you keep using the library

PR #761, opened 2025-05-12
Avoid generating `WITH` clauses for 0 CTEs

### LINKED ISSUE #760: `selectWith . pure` generates invalid SQL
Versions:

* `beam-core-0.10.1.0`
* `beam-postgres-0.5.3.1`

Code such as

```haskell
selectReturningList . selectWith . pure $ do
  all_ $ yourTable yourDatabase
```

generates invalid SQL.

A `WITH` clause requires at least 1 CTE, but `selectWith . pure` generates a query with 0 CTEs, which is a syntax error.

isaac-uptrust:
Closes https://github.com/haskell-beam/beam/issues/760

I've included the direct fix, as well as a correct-by-construction API change.
The API change isn't backwards compatible, so I understand if you want to leave it out or go through a separate process for it.

Are there any tests you'd like me to write?

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL/SQL99.hs
+++ b/beam-core/Database/Beam/Backend/SQL/SQL99.hs
@@ -16,6 +16,7 @@ module Database.Beam.Backend.SQL.SQL99
 import Database.Beam.Backend.SQL.SQL92
 
 import Data.Kind ( Type )
+import Data.List.NonEmpty ( NonEmpty )
 import Data.Text ( Text )
 
 class IsSql92SelectSyntax select =>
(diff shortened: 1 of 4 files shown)


LaurentRDC:
This is great, thanks @isaac-uptrust !

Ideally, in this PR, we could have:

* The direct fix;
* A very simple test case that would fail on master right now, but that is fixed with your patch.

I can release this very quickly. Then, could you factor out the correct-by-construction patch? We're hoping to make a release with breaking changes relatively soon.

LaurentRDC:
Superceded by #814

Issue #763, opened 2025-05-27
beam-postgres-0.5.4.2 failed to build with ghc 9.6.6


xiaoshihou514:
`ghc --version`:
```
The Glorious Glasgow Haskell Compilation System, version 9.6.6
```

build output:
```
Resolving dependencies...
Build profile: -w ghc-9.6.6 -O1
In order, the following will be built (use -v for more details):
 - beam-postgres-0.5.4.2 (lib) (requires build)
 - deepsleep-0.1.0.0 (exe:deepsleep) (first run)
Starting     beam-postgres-0.5.4.2 (lib)
Building     beam-postgres-0.5.4.2 (lib)

Failed to build beam-postgres-0.5.4.2.
Build log (
/home/xiaoshihou/.cache/cabal/logs/ghc-9.6.6/beam-postgres-0.5.4.2-3b4da6ecb8dc9be49778740aa94fac281b4ac1374bd98637cb2c266906a29e85.log
):
Configuring library for beam-postgres-0.5.4.2..
Preprocessing library for beam-postgres-0.5.4.2..
Building library for beam-postgres-0.5.4.2..
[ 1 of 14] Compiling Database.Beam.Postgres.Syntax ( Database/Beam/Postgres/Syntax.hs, dist/build/Database/Beam/Postgres/Syntax.o, dist/build/Database/Beam/Postgres/Syntax.dyn_o )
[ 2 of 14] Compiling Database.Beam.Postgres.Types ( Database/Beam/Postgres/Types.hs, dist/build/Database/Beam/Postgres/Types.o, dist/build/Database/Beam/Postgres/Types.dyn_o )
[ 3 of 14] Compiling Database.Beam.Postgres.PgSpecific ( Database/Beam/Postgres/PgSpecific.hs, dist/build/Database/Beam/Postgres/PgSpecific.o, dist/build/Database/Beam/Postgres/PgSpecific.dyn_o )
[ 4 of 14] Compiling Database.Beam.Postgres.Full ( Database/Beam/Postgres/Full.hs, dist/build/Database/Beam/Postgres/Full.o, dist/build/Database/Beam/Postgres/Full.dyn_o )
[ 5 of 14] Compiling Database.Beam.Postgres.Extensions ( Database/Beam/Postgres/Extensions.hs, dist/build/Database/Beam/Postgres/Extensions.o, dist/build/Database/Beam/Postgres/Extensions.dyn_o )
[...]

PR #764, opened 2025-06-10
beam-sqlite.cabal: skip incompatible beam-migrate

benjamin-thomas:
Hello!

Looking at beam for the first time, I struggled a bit to get going due to a dependency conflict.

The latest beam-sqlite v=0.5.4.0 is not compatible with the latest beam-migrate v=0.5.3.2

v0.5.3.2 introduced a requirement on Text vs String so a compilation error ensues.

I suppose leaving the PR open to document the minimal working setup until this is fixed is a good option too.

```
-- myProject.cabal

-- The latest beam-sqlite v=0.5.4.0 is not compatible with the latest beam-migrate v=0.5.3.2 (which it depends on)
build-depends:    base ^>=4.18.3.0
                , beam-core
                , beam-sqlite == 0.5.4.0
                , beam-migrate == 0.5.3.1
               ...
```

I didn't find it very intuitive that beam-sqlite depends on beam-migrate (I'm more interested in beam for it's query ability, not so much the migration part)

Maybe you should introduce semantic versioning?

### DIFF
--- a/beam-sqlite/beam-sqlite.cabal
+++ b/beam-sqlite/beam-sqlite.cabal
@@ -26,7 +26,7 @@ library
   build-depends:      base          >=4.11 && <5,
 
                       beam-core     >=0.10 && <0.11,
-                      beam-migrate  >=0.5  && <0.6,
+                      beam-migrate  >=0.5  && <0.5.3.2,
 
                       sqlite-simple >=0.4  && <0.5,
                       text          >=1.0  && <2.2,


LaurentRDC:
The problem is related to #763 ; when I created the previous release of beam-sqlite, the source code I uploaded to Hackage was not what I was expecting.

I'll create a new release and deprecate beam-sqlite 0.5.4.0.

About your other points:

> I didn't find it very intuitive that beam-sqlite depends on beam-migrate (I'm more interested in beam for it's query ability, not so much the migration part)

Beam's design is special amongst Haskell ORMs in that it supports multiple backends, each with possibly custom additions. The backend can support querying, but also migration support.

In principle, querying and migrating can be separated. But in practice, each backend (SQlite, Postgres, and MySQL) is made up of a single package which includes support for querying and migrating.

I am open to reviewing contributions to separate these functionality in distinct packages.

> Maybe you should introduce semantic versioning?

We currently use semantic versioning (specifically the [Haskell PVP](https://pvp.haskell.org/)). The issue at hand here is that the release process broke.

I should add a CI step to check that the release artifacts are valid -- and ideally, releases should be done through Github Actions.

benjamin-thomas:
> I am open to reviewing contributions to separate these functionality in distinct packages.

Alright, I'm taking a mental note. I realize I may have come across as being rude, so apologies if that's the case.

> The issue at hand here is that the release process broke.

Perfectly understandable then, thanks for pointing me to the PVP spec

LaurentRDC:
Don't worry, you didn't come off as rude.

I have created a new release, [beam-sqlite 0.5.4.1](https://hackage.haskell.org/package/beam-sqlite-0.5.4.1). Don't forget to update your Hackage index (`cabal update`).

If things work, please close this PR. I will be updating the CI pipeline to build release artifacts separately

PR #770, opened 2025-08-09
Postgres Array functions

kushagarr:
Add Postgres array functions: append, prepend, remove, replace, shuffle, sample, and to_string with tests

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/PgSpecific.hs
+++ b/beam-postgres/Database/Beam/Postgres/PgSpecific.hs
@@ -161,6 +161,7 @@ import qualified Database.PostgreSQL.Simple.Range as Pg
 
 import           GHC.TypeLits
 import           GHC.Exts hiding (toList)
+import           Data.Text (Text)
 
 -- ** Postgres-specific functions
 
@@ -489,9 +490,9 @@ arrayRemove_ (QExpr arr) (QExpr el) =
 -- -- => {1,2,3,4}
 -- @
 arrayReplace_
-  :: QGenExpr ctxt Postgres s (V.Vector a)
-  -> QGenExpr ctxt Postgres s a
-  -> QGenExpr ctxt Postgres s a
+  :: QGenExpr ctxt Postgres s (V.Vector a) -- ^ The array to operate on
+  -> QGenExpr ctxt Postgres s a             -- ^ The value to be replaced
+  -> QGenExpr ctxt Postgres s a             -- ^ The new value
   -> QGenExpr ctxt Postgres s (V.Vector a)
 arrayReplace_ (QExpr arr) (QExpr fromVal) (QExpr toVal) =
   QExpr (PgExpressionSyntax . mappend (emit "array_replace") . pgParens . mconcat <$> sequenceA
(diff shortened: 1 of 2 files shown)


kushagarr:
@LaurentRDC  Could you guide me here a bit.
For some reason, array_to_string is not being formed correctly, but I am trying to get it to print the actual sql which is getting formed but it is not showing when I run the tests even if I use runBeamPostgresDebug putStrLn.

How can I get the tests to print the malformed sql and for that matter all the function sqls which I have tried to add to the repo.
Please help.

kushagarr:
Never mind I figured out the issue, the test name cannot have space.
all tests should pass now

beam-postgres/Database/Beam/Postgres/PgSpecific.hs
```diff
@@ -410,6 +413,147 @@ isSubsetOf_ (QExpr needles) (QExpr haystack) =
...
+--
+-- Example:
+--
+-- @
+-- select_ $ pure $ arrayReplace_ (val_ $ V.fromList [1::Int32,2,5,4]) (val_ 5) (val_ 3)
+-- -- => {1,2,3,4}
+-- @
+arrayReplace_
+  :: QGenExpr ctxt Postgres s (V.Vector a)
+  -> QGenExpr ctxt Postgres s a
```
LaurentRDC, line 493: Could we add a docstring to show which element will be replaced, and which is the replacement? Something like

```haskell
arrayReplace_ 
    :: QGenExpr ctxt Postgres s (V.Vector a)
    -- | Element to be replaced
    -> QGenExpr ctxt Postgres s a
    -- | Replacement element
    -> QGenExpr ctxt Postgres s a
    -> QGenExpr ctxt Postgres s (V.Vector a)
```

I know this information is also present in the example, but having parameter docstrings is faster to read

beam-postgres/Database/Beam/Postgres/PgSpecific.hs
```diff
@@ -410,6 +413,147 @@ isSubsetOf_ (QExpr needles) (QExpr haystack) =
...
+arraySample_
+  :: Integral n
+  => QGenExpr ctxt Postgres s (V.Vector a)
+  -> QGenExpr ctxt Postgres s n
+  -> QGenExpr ctxt Postgres s (V.Vector a)
+arraySample_ (QExpr arr) (QExpr n) =
+  QExpr (PgExpressionSyntax . mappend (emit "array_sample") . pgParens . mconcat <$> sequenceA
+    [ fromPgExpression <$> arr
+    , pure (emit ", ")
+    , fromPgExpression <$> n
```
LaurentRDC, line 527: What happens if `n` is 0 or negative? It's not mentioned in the [PostgreSQL documentation](https://www.postgresql.org/docs/current/functions-array.html).

In any case, how about we use `max 0 n` instead of `n`, and mention in the documentation that negative `n` has no effect here

beam-postgres/Database/Beam/Postgres/PgSpecific.hs
```diff
@@ -410,6 +413,147 @@ isSubsetOf_ (QExpr needles) (QExpr haystack) =
...
+    ])
+
+-- | Postgres array_to_string(array, delimiter) function.
+-- Converts each element to text and joins with the delimiter. NULLs are omitted.
+arrayToString_
+  :: QGenExpr ctxt Postgres s (V.Vector a)
+  -> QGenExpr ctxt Postgres s text
```
LaurentRDC, line 534: Why is the `text` type variable not `Text`? The [PostgreSQL documentation](https://www.postgresql.org/docs/current/functions-array.html) makes it clear that the delimiter is `TEXT`, and the result is `TEXT` also. I would expect:

```haskell
arrayToString_
  :: QGenExpr ctxt Postgres s (V.Vector a)
  -> QGenExpr ctxt Postgres s Text
  -> QGenExpr ctxt Postgres s Text
```

beam-postgres/Database/Beam/Postgres/PgSpecific.hs
```diff
@@ -410,6 +413,147 @@ isSubsetOf_ (QExpr needles) (QExpr haystack) =
...
+    delimExpr <- fromPgExpression <$> delim
+    pure $ emit "array_to_string" <> pgParens (arrExpr <> emit ", " <> delimExpr))
+
+-- | Postgres array_to_string(array, delimiter, null_string) function.
+-- Converts each element to text and joins with the delimiter. NULLs are
+-- represented by the provided @null_string@.
+arrayToStringWithNull_
+  :: QGenExpr ctxt Postgres s (V.Vector a)
+  -> QGenExpr ctxt Postgres s text
+  -> QGenExpr ctxt Postgres s text
```
LaurentRDC, line 548: Like my [other comment](https://github.com/haskell-beam/beam/pull/770/files#r2265383721), I think the type signature should use the `Text` type directly

LaurentRDC (changes requested):
Thanks for your contributions!

All my comments are geared to prevent users from misusing the new functions. Let me know what you think

LaurentRDC:
[Released in 0.5.4.4](https://hackage.haskell.org/package/beam-postgres-0.5.4.4)!

PR #774, opened 2025-10-16
Proper use of SQLite's `RETURNING` mechanism for inserts

### LINKED ISSUE #773: SQLite INSERT ON CONFLICT RETURNING does not return conflicts
With the SQLite backend, it seems that  `runInsertReturningList $ insertOnConflict ...` does not work correctly. It will return the newly inserted entries into the table but it does not return any of the entries that were updated on conflict. That seems like a footgun to me although I am very far from being a database expert.

<details>
<summary>Show/hide reproducer</summary>

```hs

{-# LANGUAGE DeriveAnyClass #-}
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE DerivingStrategies #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE StandaloneDeriving #-}
{-# LANGUAGE TypeFamilies #-}

module Main where

-- base
import Data.Int
  ( Int32 )

-- text
import Data.Text
  ( Text )

-- beam-core
import Database.Beam
import Database.Beam.Backend.SQL.BeamExtensions
  ( onConflictUpdateSetWhere
  , insertOnConflict
  , conflictingFields
  , runInsertReturningList
  )

-- beam-migrate
import Database.Beam.Migrate
  ( defaultMigratableDbSettings )
import Database.Beam.Migrate.Simple
  ( CheckedDatabaseSettings, autoMigrate )

-- beam-sqlite
import Database.Beam.Sqlite
  ( Sqlite, runBeamSqliteDebug )
import Database.Beam.Sqlite.Migrate
  ( migrationBackend )

-- sqlite-simple
import Database.SQLite.Simple
  ( open )

--------------------------------------------------------------------------------

data TestDb f
    = TestDb
    { usersTable :: f (TableEntity User)
    }
    deriving stock Generic

deriving anyclass instance Database be TestDb

testDb :: DatabaseSettings be TestDb
testDb = defaultDbSettings

checkedDb :: CheckedDatabaseSettings Sqlite TestDb
checkedDb = defaultMigratableDbSettings

data User f
    = User
    { userId   :: Columnar f Int32
    , userName :: Columnar f Text
    }
    deriving stock Generic

deriving stock instance Show ( PrimaryKey User Identity )
deriving stock instance Show ( User Identity )

instance Table User where
  newtype PrimaryKey User f = UserId ( Columnar f Int32 )
    deriving stock Generic
  primaryKey = UserId . userId

deriving anyclass instance Beamable User
deriving anyclass instance Beamable ( PrimaryKey User )

main :: IO ()
main = do

  conn <- open "test.sqlite"

  conflicts <-
    runBeamSqliteDebug putStrLn conn $ do

      -- Use beam-migrate to create the db for simplicity
      autoMigrate migrationBackend checkedDb

      -- Insert some users
      runInsert $
        insert ( usersTable testDb )
          ( insertValues [ User 0 "user0", User 2 "user2", User 5 "user5" ] )

      let newUsers = [ User 1 "user1", User 2 "different_user2" ]

      -- Insert some conflicting users
      runInsertReturningList $
        insertOnConflict ( usersTable testDb )
          ( insertValues newUsers )
          ( conflictingFields userId )
          ( onConflictUpdateSetWhere
            ( \ ( User { userName = fld  } )
                ( User { userName = excl } ) ->
                  fld <-. excl
            )
            ( \ ( User { userName = fld  } )
              ( User { userName = excl } ) ->
                current_ fld /=. excl
            )
          )

  print conflicts
```
</details>

This returns the conflicts as `[User {userId = 1, userName = "user1"}]`. `User 1` is newly added so that's valid, but it's missing the conflicting `User 2`. Of course the new `User 2` with name `different_user2` is properly updated in the database, overwriting the old `User 2`, but it's not returned as a conflict.

I think this may be due to the way that `beam-sqlite` [emulates `runInsertReturningList`](https://hackage-content.haskell.org/package/beam-sqlite-0.5.4.1/docs/src/Database.Beam.Sqlite.Connection.html#runInsertReturningList). I think it would be good to get rid of that code and to instead use `RETURNING`, as that is [supported since SQLite version 3.35.0 (2021-03-12)](https://www.sqlite.org/lang_returning.html). That should also enable support for `DELETE RETURNING` in the SQLite backend, which would be quite useful as well.

LaurentRDC:
The behavior of `runInsertReturningList` currently fails to return rows which had a conflict on insert. This is indirectly because `beam-sqlite` emulates the behavior of INSERT .. ON CONFLICT ... RETURNING.

SQLite 3.35 added support for a proper `RETURNING` clause on inserts. We should use this instead.

Fixes #773

### DIFF
--- a/beam-sqlite/ChangeLog.md
+++ b/beam-sqlite/ChangeLog.md
@@ -1,8 +1,17 @@
 # Unreleased
 
+## Added features
+
+* `runInsertReturningList` now uses SQLite's relatively new `RETURNING` clause.
+
+## Bux fixes
+
+* Fixed an issue where values inserted with conflicts did not return then when using `runInsertReturningList` (#774) 
+
 ## Updated dependencies
 
-* Updated the upper bound on `time` to include `time-1.14`
+* Updated the lower bound of `direct-sqlite` to `2.3.27`.
+* Updated the upper bound on `time` to include `time-1.14`.
 
 # 0.5.4.1
 


beam-sqlite/Database/Beam/Sqlite/Connection.hs
```diff
@@ -71,25 +70,24 @@ import           Data.ByteString.Builder (toLazyByteString)
...
+                           , ZonedTime, utc, utcToLocalTime )
 import           Data.Typeable (cast)
 import           Data.Word
 import           GHC.TypeLits
 
 import           Network.URI
 
 import           Text.Read (readMaybe)
+import Data.IORef (newIORef, atomicModifyIORef')
+import GHC.IORef (atomicModifyIORef'_)
```
sheaf: This doesn't seem to fit the style of the existing import list.

beam-sqlite/test/Database/Beam/Sqlite/Test/InsertOnConflictReturning.hs
```diff
@@ -0,0 +1,133 @@
...
+
+{-
+    This module is based on the bug reproducer from issue #773
+    https://github.com/haskell-beam/beam/issues/773
+-}
+module Database.Beam.Sqlite.Test.InsertOnConflictReturning (tests) where
+
+import Data.Int (Int32)
+import Data.Text (Text)
+import Data.Time (Day (..), LocalTime, UTCTime (..), fromGregorian, getCurrentTime, secondsToDiffTime)
```
sheaf: I don't think this import is used, is it?

beam-sqlite/Database/Beam/Sqlite/Connection.hs
sheaf: The comment `-- * Emulated @INSERT RETURNING@ support` should be updated as well (GitHub didn't let me include that line in the review).

sheaf:
Fantastic, I'm really happy to see this.

I also looked into adding support for `DELETE ... RETURNING`, the following seems to work to return entire rows:

```hs
instance Beam.MonadBeamDeleteReturning Sqlite SqliteM where
  runDeleteReturningList = runDeleteReturningList

runDeleteReturningList :: (Beamable table, FromBackendRow Sqlite (table Identity))
                       => SqlDelete Sqlite table
                       -> SqliteM [ table Identity ]
runDeleteReturningList (SqlDelete _ deleteCommand) =
  runReturningList $ SqliteCommandSyntax $
    fromSqliteDelete deleteCommand <> emit " RETURNING *"
```

I don't understand how we are supposed to choose what to return in a "delete returning list", e.g. if we want to only return the ids of deleted rows for instance.

Your patch is working great on simple examples, but I will let you know how it goes on a more complicate database setup later today. (What I have is still pretty simple though, so it's not gonna be a great correctness stress test, especially as I don't use default values much.)

beam-sqlite/Database/Beam/Sqlite/Connection.hs
```diff
@@ -335,11 +333,64 @@ instance MonadBeam Sqlite SqliteM where
...
+      -- incrementally to 'action'.
+      SqliteM $ do
+        (logger, conn) <- ask
+        resultsRef <- liftIO (newIORef [])
+        forM_ es $ \row -> do
+          -- RETURNING is only supported by SQLite 3.35+, which requires direct-sqlite 2.3.27+
+          let returningClause = emit " RETURNING " <> commas (map quotedIdentifier fields)
+              (insertFields, insertRow) = unzip $ filter ((/= SqliteExpressionDefault) . snd) $ zip fields row
+              SqliteSyntax cmd vals = formatSqliteInsertOnConflict tbl insertFields (SqliteInsertExpressions [ insertRow ]) onConflict <> returningClause
+              cmdString = BL.unpack (toLazyByteString (withPlaceholders cmd))
```
sheaf: Only a minor comment (that probably should be punted to a separate ticket), but: would it be possible to instead classify the rows by which fields have DEFAULT values? For example, if all the rows to be inserted have DEFAULT in the same position, then we don't need to do this, right?

sheaf:
The following works for "DELETE ... RETURNING":


```hs
newtype SqliteDeleteReturning a = SqliteDeleteReturning SqliteDeleteSyntax

data SqliteInaccessible

runDeleteReturningList
  :: ( MonadBeam be m
     , BeamSqlBackendSyntax be ~ SqliteCommandSyntax
     , FromBackendRow be a
     )
  => SqliteDeleteReturning a
  -> m [a]
runDeleteReturningList (SqliteDeleteReturning (SqliteDeleteSyntax syntax)) =
  runReturningList $ SqliteCommandSyntax syntax

-- | SQLite @DELETE ... RETURNING@ statement support. The last
-- argument takes the newly inserted row and returns the values to be
-- returned. Use 'runDeleteReturningList' to get the results.
deleteReturning :: forall a table be. Projectible Sqlite a
                => DatabaseEntity Sqlite be (TableEntity table)
                -> (forall s. table (QExpr Sqlite s) -> QExpr Sqlite s Bool)
                -> (table (QExpr Sqlite SqliteInaccessible) -> a)
                -> SqliteDeleteReturning (QExprToIdentity a)
deleteReturning table@(DatabaseEntity (DatabaseTable { dbTableSettings = tblSettings }))
                mkWhere
                mkProjection =
  SqliteDeleteReturning $ SqliteDeleteSyntax $
    fromSqliteDelete sqliteDelete
      <>
    emit " RETURNING " <> commas (map fromSqliteExpression (project (Proxy @Sqlite) (mkProjection tblQ) "t"))
  where

    SqlDelete _ sqliteDelete = delete table mkWhere
    tblQ = changeBeamRep getFieldName tblSettings
    getFieldName (Columnar' fd) =
      Columnar' $ QExpr $ pure $ fieldE (unqualifiedField (_fieldName fd))
```

I think the monads such as `MonadBeamDeleteReturning` etc are incorrectly designed because their type signatures all return the entire table while one might want to project out only certain parts of the table. This is why the Postgres backend defines functions such as `runPgDeleteReturningList` where the return type is in terms of `a` (the thing we are returning) and not in terms of the original `table`.

sheaf:
Here's my attempt to group all the insertions by the collection of defaults that should be omitted from the query:

```hs
newtype SqliteInsertReturning a = SqliteInsertReturning [SqliteSyntax]

runInsertReturningList
  :: ( MonadBeam be m
     , BeamSqlBackendSyntax be ~ SqliteCommandSyntax
     , FromBackendRow be a
     )
  => SqliteInsertReturning a
  -> m [a]
runInsertReturningList (SqliteInsertReturning es) =
  concat <$>
    traverse ( \ syntax -> runReturningList $ SqliteCommandSyntax syntax ) es

data SqliteInaccessible

insertOnConflictReturning
  :: forall s a table db. Projectible Sqlite a
  => DatabaseEntity Sqlite db (TableEntity table)
  -> SqlInsertValues Sqlite (table (QExpr Sqlite s))
  -> SqlConflictTarget Sqlite table
  -> SqlConflictAction Sqlite table
  -> (table (QExpr Sqlite SqliteInaccessible) -> a)
        -- ^ projection describing what to return
  -> SqliteInsertReturning (QExprToIdentity a)
insertOnConflictReturning
  table@(DatabaseEntity (DatabaseTable { dbTableSettings = tblSettings }))
  vals tgt action mkProjection =
  SqliteInsertReturning $
    case insertOnConflict table vals tgt action of
      SqlInsert _ (SqliteInsertSyntax tbl fields values onConflict) ->
        [ formatSqliteInsertOnConflict tbl fields' values' onConflict <>
          emit " RETURNING " <> commas (map fromSqliteExpression (project (Proxy @Sqlite) (mkProjection tblQ) "t"))
        | ( fields', values' ) <- sqliteGroupByDefaults fields values
        ]
      SqlInsertNoRows -> []
  where
    tblQ = changeBeamRep getFieldName tblSettings
    getFieldName (Columnar' fd) =
      Columnar' $ QExpr $ pure $ fieldE (unqualifiedField (_fieldName fd))

-- | SQLite cannot deal with DEFAULT in insertion. Instead, it expects these
-- fields to be omitted from the INSERT.
--
-- This means that, for any given INSERT query, the row value being inserted
-- must have a fixed set of fields with DEFAULT value.
--
-- This function groups the inserted values by the set of DEFAULT fields, so
-- that we can preform a separate INSERT for each set of DEFAULT fields.
sqliteGroupByDefaults :: [ Text ] -> SqliteInsertValuesSyntax -> [([Text], SqliteInsertValuesSyntax)]
sqliteGroupByDefaults flds ( SqliteInsertExpressions es ) =
  let
    partitionDefaultsFrom :: (IntSet, [SqliteExpressionSyntax]) -> Int -> [SqliteExpressionSyntax] -> (IntSet, [SqliteExpressionSyntax])
    partitionDefaultsFrom (defs, acc) _ [] = ( defs, reverse acc )
    partitionDefaultsFrom (defs, acc) i (x:xs)
      | x == SqliteExpressionDefault
      = partitionDefaultsFrom ( IntSet.insert i defs, acc ) ( i + 1 ) xs
      | otherwise
      = partitionDefaultsFrom ( defs, x : acc ) ( i + 1 ) xs

    grouped :: [ ( IntSet, [[SqliteExpressionSyntax]] ) ]
    grouped =
      Strict.Map.toList $
      Strict.Map.fromListWith (++)
        [ ( defaults, [ e' ] )
        | e <- es
        , let ( defaults, e' ) = partitionDefaultsFrom ( IntSet.empty, [] ) 0 e
        ]

    mkSyntax dflts xs =
      ( [ fld | ( i, fld ) <- zip [0..] flds
              , not $ i `IntSet.member` dflts
              ]
        , SqliteInsertExpressions xs )

  in
    map ( uncurry mkSyntax ) grouped
sqliteGroupByDefaults flds orig@( SqliteInsertFromSql{}) =
  -- Preserve manually-written queries
  [(flds, orig)]
  ```

beam-sqlite/Database/Beam/Sqlite/Connection.hs
LaurentRDC: Thanks! I'm used to using `ormolu` so much that I forgot about the style

beam-sqlite/test/Database/Beam/Sqlite/Test/InsertOnConflictReturning.hs
LaurentRDC: Indeed! Why did my editor not tell me?

beam-sqlite/Database/Beam/Sqlite/Connection.hs
LaurentRDC: Yes that's very true. 
I considered doing something like it, but I started with this simpler implementation instead because the order of insertion is preserved this way. If you grouped rows by the positions of their default values, then you would have to reconstruct the return order.
This isn't impossible, but this is significantly more complex. As you mention, this could be punted to a new ticket.

Actually, I see you have made a prototype implementation for this below; would you like to submit a PR?

PR #780, opened 2026-02-08
DuckDB backend

LaurentRDC:
Fixes #779


.github/workflows/build.yaml
```diff
@@ -63,6 +78,16 @@ jobs:
           ${{ steps.setup-haskell.outputs.cabal-store }}
         key: ${{ runner.os }}-${{ hashFiles('dependencies-versions', 'cabal.project.local') }}-cabal-install
 
+    - name: Install DuckDB artifact
+      if: ${{ matrix.duckdb }}
+      run: |
+        curl -L -o libduckdb.zip https://github.com/duckdb/duckdb/releases/download/v1.4.1/libduckdb-linux-amd64.zip && \
```
kmicklas: Any reason for the `&& \`s as opposed to separate commands?

docs/user-guide/backends/beam-duckdb.md
```diff
@@ -0,0 +1,183 @@
...
+querying from CSV files:
+
+```haskell
+data ExampleDB f = ExampleDB
+  { _exams :: f (CSVEntity ExamT),
+  }
+  deriving (Generic, Database DuckDB)
+```
+
+Contrary to `parquet` and `icebergTable`, we'll need to tweak the default CSV options by changing
```
kmicklas: I don't see anything wrong, but starting with this paragraph GitHub started highlighting the markdown as if it were Haskell, which makes me suspicious that there is something wrong with the code block above. (Evil Unicode?)

beam-duckdb/src/Database/Beam/DuckDB/Syntax/Extensions/CSV.hs
```diff
@@ -0,0 +1,213 @@
...
+                  <> emitOptions (csvFileOptions desc)
+                  <> emit ") AS "
+                  <> quotedIdentifier name
+          )
+          (tableFieldsToExpressions (csvTableSettings desc))
+          (const Nothing)
+          snd
+      )
+  where
+    quotePath path = mconcat [emitChar '\'', emit (Text.pack path), emitChar '\'']
```
kmicklas: Is it possible to use placeholders for paths? This seems like a possible injection mechanism.

kmicklas (commented):
Is there a reason for having separate entity types for each data source? I can't think of any extra power this would provide. On the other hand, with a single entity type you could in theory do more dynamic behavior like choose at runtime whether to read from CSV or parquet with the same schema.

.github/workflows/build.yaml
LaurentRDC: No reason other than I lifted these commands from the `duckdb-simple` library which builds inside a Dockerfile. I'll break the commands down

beam-duckdb/src/Database/Beam/DuckDB/Syntax/Extensions/CSV.hs
LaurentRDC: That is a great question. [It looks like `read_csv` and functions like it don't support placeholders](https://github.com/duckdb/duckdb/discussions/9019). As far as I understand, this means that the filepath or glob cannot be used for injection

LaurentRDC:
Well, there's a simpler design actually that removes a lot of duplicated code. I suspect this is what Ken was suggesting:

```haskell
data ExternalEntity (table :: (Type -> Type) -> Type)

data Configuration
    = CSV (NonEmpty FilePath) CSVOptions
    | Parquet (NonEmptyFilePath) -- no parquet options
    | Iceberg FilePath IcebergOptions

-- To replace `parquet`, `csv`, and `icebergTable`
externalEntity :: Configuration 
               -> EntityModification (DatabaseEntity DuckDB db) DuckDB (ExternalEntity table)
```

I will think of better names for things, but the interface itself is nice

kmicklas:
> Well, there's a simpler design actually that removes a lot of duplicated code. I suspect this is what Ken was suggesting:
> 
> ```haskell
> data DataSourceEntity (table :: (Type -> Type) -> Type)
> 
> data DataSource
>     = CSV (NonEmpty FilePath) CSVOptions
>     | Parquet (NonEmptyFilePath) -- no parquet options
>     | Iceberg FilePath IcebergOptions
> 
> dataSource :: DataSource 
>            -> EntityModification (DatabaseEntity DuckDB db) DuckDB (DataSource table)
> ```

Yup! This is exactly what I meant.

My only remaining concern is that I think we should document more clearly the injection possibility with untrusted paths.

kmicklas (approved):

Issue #727, opened 2024-11-18
Automate building and deploying documentation using GitHub Actions


LaurentRDC:
Currently, we manually deploy documentation to GitHub Pages using the `build-docs.sh` script. 

This should be automated using GitHub actions.

LaurentRDC:
@tathougies would you be able to briefly describe your process for building documentation using nix?

LaurentRDC:
Fixed by #784

PR #784, opened 2026-02-28
Modernize documentation setup

### LINKED ISSUE #612: Publish docs from CI

LaurentRDC:
This PR does two things:

* Modernize the way documentation is built by re-using the `flake.nix` mechanism for development;
* Generates documentation on every commit and deploy to Github Pages once again.

Fixes #612 and #727

## Modernization

There were a few issues that needed to be fixed.

The most important is that I could not get the `dbLenses` and `tableLenses` mechanism so compile on GHC 9.8 or 9.10. I recall there were changes to `ImpredicativeTypes` in GHC 9.8, so it could be why. Instead, I generated lenses using `microlens-th`'s `makeLenses`, and edited the section on lenses accordingly (i.e. practically removed it).

This introduced `TemplateHaskell` in some of the examples. Unfortunately, `runhaskell` doesn't work well with staged compilation of files containing `TemplateHaskell`. Therefore, we build examples with `runhaskell` UNLESS Template Haskell is involved, in which case we use the much slower `ghc --make -O0`.


LaurentRDC:
@peterbecich I extended your implementation of `flake.nix` from #737 to also work with the documentation, but I'm very far out of my depth here. Would you take a look?

PR #789, opened 2026-03-16
beam-migrate: CREATE INDEX support

sheaf:
This PR adds functionality to add secondary indices to a database to speed up queries.

The user-facing API consists of the `addTableIndex` function, and the helper `indexCol` function.

Backend support goes via the new typeclass `IsSql92CreateDropIndexSyntax`, with support in both the SQLite and Postgres backends.

### DIFF
--- a/beam-migrate/Database/Beam/Migrate/Actions.hs
+++ b/beam-migrate/Database/Beam/Migrate/Actions.hs
@@ -1,6 +1,7 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -83,6 +84,8 @@ module Database.Beam.Migrate.Actions
   , addColumnProvider
   , addColumnNullProvider
   , dropColumnNullProvider
+  , createIndexActionProvider
+  , dropIndexActionProvider
   , defaultActionProvider
   , defaultSchemaActionProvider
 
@@ -112,7 +115,6 @@ import qualified Data.Sequence as Seq
 import           Data.Text (Text)
 import qualified Data.Text as T
 import           Data.Typeable
-import           Data.Semigroup
 
 import           GHC.Generics
 
@@ -272,6 +274,10 @@ createTableWeight = 500
 dropTableWeight = 100
 addColumnWeight = 1
 dropColumnWeight = 1
+
+createIndexWeight, dropIndexWeight :: Int
+createIndexWeight = 200
+dropIndexWeight = 50
 
 -- | Proceeds only if no predicate matches the given pattern. See the
 -- implementation of 'dropTableActionProvider' for an example of usage.
@@ -516,6 +522,60 @@ dropColumnNullProvider = ActionProvider provider
          pure (PotentialAction (HS.fromList [SomeDatabasePredicate colP]) mempty
                                (Seq.singleton (MigrationCommand cmd MigrationKeepsData))
                                ("Drop not null constraint for " <> colNm <> " on " <> qnameAsText tblNm) 100)
+
+-- | Action provider for @CREATE INDEX@ actions.
+--
+-- Generates a @CREATE INDEX@ command whenever the destination schema contains
+-- a 'TableHasIndex' predicate that is not satisfied in the current state.
+createIndexActionProvider :: forall be
+                           . ( BeamMigrateOnlySqlBackend be
+                             , IsSql92UniqueIndexSyntax (BeamSqlBackendSyntax be) )
+                          => ActionProvider be
+createIndexActionProvider = ActionProvider provider
+  where
+    provider :: ActionProviderFn be
+    provider findPreConditions findPostConditions =
+      do (idxP@(TableHasIndex { hasIndex_table = postTblNm, hasIndex_name = idxNm
+                              , hasIndex_columns = idxCols, hasIndex_opts = idxOpts })
+            :: TableHasIndex be) <- findPostConditions
+         -- Ensure this index doesn't already exist
+         ensuringNot_ $
+           do (TableHasIndex { hasIndex_table = preTblNm, hasIndex_name = idxNm' }
+                :: TableHasIndex be) <- findPreConditions
+              guard (preTblNm == postTblNm && idxNm' == idxNm)
+         -- Ensure the target table already exists
+         TableExistsPredicate tblNm' <- findPreConditions
+         guard (tblNm' == postTblNm)
+
+         let cmd = createIndexCmd idxNm (qnameAsTableName postTblNm) idxCols idxOpts
+         pure (PotentialAction mempty (HS.singleton (p idxP))
+                               (Seq.singleton (MigrationCommand cmd MigrationKeepsData))
+                               ("Create index " <> idxNm <> " on " <> qnameAsText postTblNm)
+                               createIndexWeight)
+
+-- | Action provider for @DROP INDEX@ actions.
+dropIndexActionProvider :: forall be
+                         . ( BeamMigrateOnlySqlBackend be
+                           , IsSql92UniqueIndexSyntax (BeamSqlBackendSyntax be) )
+                        => ActionProvider be
+dropIndexActionProvider = ActionProvider provider
+  where
+    provider :: ActionProviderFn be
+    provider findPreConditions findPostConditions =
+      do (idxP@(TableHasIndex { hasIndex_table = preTblNm, hasIndex_name = idxNm })
+            :: TableHasIndex be) <- findPreConditions
+         ensuringNot_ $
+           do (TableHasIndex { hasIndex_table = postTblNm, hasIndex_name = idxNm' }
+                :: TableHasIndex be) <- findPostConditions
+              guard (preTblNm == postTblNm && idxNm' == idxNm)
+
+         let cmd = dropIndexCmd idxNm
+         pure (PotentialAction (HS.singleton (p idxP)) mempty
+           -- Dropping a secondary index doesn't lose data, as the index
+           -- can be recalculated.
+                               (Seq.singleton (MigrationCommand cmd MigrationKeepsData))
+                               ("Drop index " <> idxNm <> " on " <> qnameAsText preTblNm)
+                               dropIndexWeight)
 
 -- | Default action providers for any SQL92 compliant syntax.
 --
--- a/beam-migrate/Database/Beam/Migrate/Checks.hs
+++ b/beam-migrate/Database/Beam/Migrate/Checks.hs
@@ -131,7 +131,7 @@ instance ( Typeable be
          DatabasePredicate (TableHasIndex be) where
   englishDescription (TableHasIndex { hasIndex_table = tbl, hasIndex_name = nm
                                     , hasIndex_columns = cols, hasIndex_opts = opts }) =
-    (if indexIsUnique opts then "Unique index " else "Index ") <>
+    (if indexIsUnique @(BeamSqlBackendSyntax be) opts then "Unique index " else "Index ") <>
     show nm <> " on table " <> show tbl <> " covering columns " <> show cols
 
   predicateSpecificity _ = PredicateSpecificityAllBackends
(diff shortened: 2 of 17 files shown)


beam-postgres/Database/Beam/Postgres/Migrate.hs
```diff
@@ -402,14 +405,51 @@ getDbConstraintsForSchemas subschemas conn =
...
+           -- exclude primary key indices
+         , "  AND NOT ix.indisprimary"
+           -- exclude indices created implicitly by a UNIQUE or EXCLUDE constraint
+         , "  AND NOT EXISTS (SELECT 1 FROM pg_constraint con WHERE con.conindid = ix.indexrelid)"
+           -- exclude expression indices: a key column number of 0 means that
+           -- position is an expression (e.g. lower(col)) rather than a plain
+           -- column reference, which TableHasIndex cannot represent
+         , "  AND NOT EXISTS (SELECT 1 FROM unnest(ix.indkey) AS k(attnum) WHERE k.attnum = 0)"
+         , "GROUP BY ns.nspname, c.relname, i.relname, ix.indisunique" ]))
+
```
sheaf, line 452: Claude Code helped me write this, because I was completely out of my depth with it. I can't really vouch for its correctness.

LaurentRDC:
Thanks! I should be able to review in the next few days.

As a side-note: did you check out other attempts at this? e.g. https://github.com/haskell-beam/beam/pull/335

sheaf:
> As a side-note: did you check out other attempts at this? e.g. #335

I wasn't aware of this approach, I'll take a look this week.

sheaf:
#335 was adding the notion of index to `beam-core`, at a more fundamental level in the database schema, while this PR only adds it to `beam-migrate`. This means the current PR is more limited in scope, as it only really supports creating/dropping secondary indices (and computing their presence by introspection), but it doesn't include them in the schema itself.

#335 has a lot of complexity due to trying to derive indices from a separate record description of indices using generics, with a separate `DatabaseIndices` type. This allows indices to be automatically generated, but judging from PR review the approach had some fundamental limitations.  
In this PR the approach is more low-level, not far off from just declaring the indices as simple field names (but using the `indexCol` function to retrieve the names instead of manually writing them). 

#335 was missing SQLite support, and had no schema introspection queries.

#335 also made `DROP INDEX` into a sub-command of `ALTER TABLE`, which I don't think is correct.

I took a couple of changes from that PR and pushed them here:

  - use `MigrationKeepsData` for `DROP INDEX` because it doesn't actually lose data in the database (the indices can be recalculated)
  - introduce an `IndexOptions` datatype instead of raw `Bool` for uniqueness, to make the design more extensible

sheaf:
In conclusion, I think this is simpler than #335 because it avoids the machinery for deriving secondary indices using Generics. That part also seemed to be what caused #335 to get bogged down. With this PR the approach is a bit more manual (in particular, the secondary indices are named manually).

None of this is meant as a judgement on #335 because that functionality is indeed quite appealing.

My use case is mainly to automate away manual index creation and have the migrations framework handle it, which this PR does quite well. But otherwise the approach here is a bit barebones in comparison.

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
```diff
@@ -240,3 +242,38 @@ class Sql92SerializableDataTypeSyntax dataType where
...
+-- indexOptions { indexUnique = True }
+-- @
+data IndexOptions = IndexOptions
+  { indexUnique :: Bool  -- ^ Whether the index enforces uniqueness (@UNIQUE@)
+  } deriving (Show, Eq, Ord, Generic)
+instance Hashable IndexOptions
+
+-- | Default index options: a non-unique index.
+indexOptions :: IndexOptions
+indexOptions = IndexOptions { indexUnique = False }
```
LaurentRDC: Can we make `IndexOptions` polymorphic over the backend? Probably a type family inside of `IsSql92CreateDropIndexSyntax`

I'm thinking of PostgreSQL indices which [have a lot of configuration](https://www.postgresql.org/docs/current/sql-createindex.html). Supporting all of this is asinine, but supporting e.g. `UNIQUE` and `NULLS DISTINCT` should be at least doable.

beam-migrate/Database/Beam/Migrate/Actions.hs
```diff
@@ -112,7 +114,9 @@ import qualified Data.Sequence as Seq
 import           Data.Text (Text)
 import qualified Data.Text as T
 import           Data.Typeable
-import           Data.Semigroup
+# if !MIN_VERSION_base(4,11,0)
+import Data.Semigroup
+#endif
```
LaurentRDC: We can get rid of this since the oldest supported GHC is 8.10, and base 4.11 is much older I think

LaurentRDC (commented):
Looking pretty good!

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
tathougies: The normal approach I've taken (no need to follow though if people disagree) is to use the 'finally tagless' style. This means, creating a type family inside `IsSql92CreateDropIndexSyntax` like `Sql92CreateDropIndexUniqueSyntax` and then give a type class like `IsSql92CreateDropIndexUniqueSyntax` that this indexed type family ought to satisfy. Then, future constructors can be added by subclassing that class.

For a full treatment see Kiselyov's notes: https://okmij.org/ftp/tagless-final/course/lecture.pdf

beam-migrate/Database/Beam/Migrate/Actions.hs
sheaf: Ah, good to know. I had just copied that from elsewhere; I'm updating to remove the other conditions as well.

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
sheaf: I have implemented this change now, including the additional typeclass for index uniqueness constraints.

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
```diff
@@ -240,3 +238,54 @@ class Sql92SerializableDataTypeSyntax dataType where
...
+-- | Syntax extension for @CREATE INDEX@ and @DROP INDEX@ DDL commands.
+--
+-- @CREATE INDEX@ is not part of SQL92 proper, but is a widely supported
+-- extension.
+class ( IsSql92DdlCommandSyntax syntax
+      , Show     (Sql92CreateIndexOptionsSyntax syntax)
+      , Eq       (Sql92CreateIndexOptionsSyntax syntax)
+      , Hashable (Sql92CreateIndexOptionsSyntax syntax)
+      ) => IsSql92CreateDropIndexSyntax syntax where
+  data family Sql92CreateIndexOptionsSyntax syntax
```
LaurentRDC: I'm not familiar with data families. I would have expected a closed type family instead:

```haskell
class ( IsSql92DdlCommandSyntax syntax
      , Show     (Sql92CreateIndexOptionsSyntax syntax)
      , Eq       (Sql92CreateIndexOptionsSyntax syntax)
      , Hashable (Sql92CreateIndexOptionsSyntax syntax)
      ) => IsSql92CreateDropIndexSyntax syntax where
  type Sql92CreateIndexOptionsSyntax syntax
```

That's how other syntaxes are represented in Beam.

Is there an advantage to using data families?

beam-migrate/Database/Beam/Migrate/Checks.hs
```diff
@@ -111,6 +112,39 @@ instance ( Typeable be, BeamMigrateOnlySqlBackend be
...
     | otherwise = False
 
+-- | Asserts that the given table has a secondary index with the given name
+-- covering the given columns (in order). Create these predicates with
+-- 'Database.Beam.Migrate.Types.CheckedEntities.addTableIndex'.
+data TableHasIndex be
+  = TableHasIndex
+  { hasIndex_table   :: QualifiedName                -- ^ table name
+  , hasIndex_name    :: Text                         -- ^ index name
+  , hasIndex_columns :: [Text]                       -- ^ ordered column names
```
LaurentRDC: I think this should be a non-empty list

beam-postgres/Database/Beam/Postgres/Migrate.hs
LaurentRDC: [Based on a cursory reading of the documentation it looks fine](https://www.postgresql.org/docs/current/catalog-pg-index.html). It's exercised in tests

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
sheaf: It's one less indirection. With this setup one writes:

```hs
instance IsSql92CreateDropIndexSyntax MySyntax where
  data Sql92CreateIndexOptionsSyntax MySyntax
       = MySyntaxIndexOptions { field1 :: Ty1, field2 :: Ty2 }
    deriving stock (Eq, Ord, Generic)
    deriving anyclass Hashable
```

whereas with a type family it would be:

```hs
instance IsSql92CreateDropIndexSyntax MySyntax where
  type Sql92CreateIndexOptionsSyntax MySyntax = MySyntaxIndexOptions

data MySyntaxIndexOptions
  = MySyntaxIndexOptions { field1 :: Ty1, field2 :: Ty2 }
    deriving stock (Eq, Ord, Generic)
    deriving anyclass Hashable
```

The latter is strictly more boilerplate, and also less permissive as one cannot write unsaturated type families while one can write unsaturated data families. Perhaps not so relevant here, but it does make some type-level programming idioms impossible, e.g. (rough sketch):

```hs
type KnownSyntaxes = [PgCommandSyntax, SqliteCommandSyntax]

type AllIndicesSupport :: (Type -> Constraint) -> Constraint
type AllIndicesSupport c = All (c . Sql92CreateIndexOptionsSyntax) KnownSyntaxes
```

All that said, if you think it would be better for consistency I can switch the code to using type families.

beam-migrate/Database/Beam/Migrate/Checks.hs
sheaf: I have updated the API to use non-empty lists.

sheaf:
The failures on 9.12 and 9.14 seem spurious as they have to do with the installation of `alex`/`happy` in the CI environment. 9.14 works fine for me locally.

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
LaurentRDC: In the type family case, `MySyntaxIndexOptions` has a standalone type, whereas the data family does not. Can the data family instance be used in a standalone way?

Imagine a backend-specific function on an index. Something like:

```haskell
someFunc :: PgSyntaxIndexOptions -> Pg ()
```

In the data family case, would that be written like this?

```haskell
someFunc :: Sql92CreateIndexOptionsSyntax PgSyntax -> Pg ()
```

If so, I lean towards keeping the consistency with other bits of beam by using type families

sheaf: Yes, it would be written as you say.

LaurentRDC: Allright, let's keep the consistency by using a type family instead of `data family` and then we can wrap this PR up! 
Thank you for your patience through this review process

sheaf: I've updated the class to use a type family. It has the unfortunate effect of being more ambiguous, because a type signature such as

```hs
indexIsUnique :: Sql92CreateIndexOptionsSyntax syntax -> Bool
```

which used to be unambiguous is now ambiguous (because `syntax` only appears guarded under a type family).

If you look at the commit I think you'll agree things are quite a bit less ergonomic like this, but I agree it's also important to keep the interface consistent.

LaurentRDC (approved):

LaurentRDC:
Awesome @sheaf ! Thanks for your contribution

LaurentRDC:
I should be able to make new releases for beam-migrate / beam-sqlite / beam-postgres today

LaurentRDC:
Released:
* [beam-migrate-0.5.4.0](https://hackage.haskell.org/package/beam-migrate-0.5.4.0)
* [beam-postgres-0.5.5.0](https://hackage.haskell.org/package/beam-postgres-0.5.5.0)
* [beam-sqlite-0.5.7.0](https://hackage.haskell.org/package/beam-sqlite-0.5.7.0)

PR #791, opened 2026-04-04
Add support for foreign key constraints and foreign key actions

sheaf:
This PR adds support for foreign key constraints and actions to `beam-migrate`, with support in the Postgres and SQLite backends.

  - The `ForeignKeyAction` datatype is used to represent possible actions to perform when updating or deleting a row with referencing foreign keys.
  - `createTableActionProvider` now takes an additional `ForeignKeySupport` argument: a witness of evidence for `IsSql92ForeignKeyTableConstraintSyntax` for backends that support it.
  - Postgres and SQLite support for parsing existing foreign key constraints.
  - The function `addTableForeignKey` can be used to declare new foreign key constraints.

### DIFF
--- a/beam-migrate/Database/Beam/Migrate/Actions.hs
+++ b/beam-migrate/Database/Beam/Migrate/Actions.hs
@@ -107,6 +107,7 @@ import           Control.Monad
 import           Control.Parallel.Strategies
 
 import           Data.Foldable
+import qualified Data.List.NonEmpty as NE
 
 import qualified Data.HashMap.Strict as HM
 import qualified Data.HashSet as HS
@@ -377,9 +378,41 @@ createTableActionProvider =
            guard (tblNm == postTblNm)
            pure (primaryKeyP, primaryKey)
 
-         let postConditions = [ p tblP, p primaryKeyP ] ++ concat columnsP
+         -- For each foreign key constraint, ensure the referenced table
+         -- already exists in the current state.
+         -- This enforces a topological ordering, so that the solver schedules
+         -- CREATE TABLE for all referenced tables first.
+         --
+         -- Note: this does not support cycles in the foreign key reference graph,
+         -- for which we would need ALTER TABLE ADD CONSTRAINT, which we don't
+         -- currently support.
+         (fkPs, fkConstraints) <- do
+           let fkData =
+                 unzip
+                   [ ( p fkP
+                     , foreignKeyConstraintSyntax localCols (qnameAsText refTbl) refCols onUpd onDel )
+                   | fkP@(TableHasForeignKey tblNm localCols refTbl refCols onUpd onDel)
+                       <- findPostConditions
+                   , tblNm == postTblNm
+                   ]
+               refTbls =
+                 [ refTbl
+                 | TableHasForeignKey tblNm _ refTbl _ _ _ <- findPostConditions
+                 , tblNm == postTblNm
+                 , refTbl /= postTblNm
+                 ]
+           for_ refTbls $ \refTbl -> do
+             TableExistsPredicate nm <- findPreConditions
+             guard (nm == refTbl)
+           pure fkData
+
+
+         let postConditions = [ p tblP, p primaryKeyP ] ++ concat columnsP ++ fkPs
              cmd = createTableCmd (createTableSyntax Nothing (qnameAsTableName postTblNm) colsSyntax tblConstraints)
-             tblConstraints = if null primaryKey then [] else [ primaryKeyConstraintSyntax primaryKey ]
+             tblConstraints = (case NE.nonEmpty primaryKey of
+                                Nothing   -> []
+                                Just pkey -> [primaryKeyConstraintSyntax pkey])
+                           ++ fkConstraints
              colsSyntax = map (\(colNm, type_, cs) -> (colNm, columnSchemaSyntax type_ Nothing cs Nothing)) columns
          pure (PotentialAction mempty (HS.fromList postConditions)
                                (Seq.singleton (MigrationCommand cmd MigrationKeepsData))
--- a/beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
+++ b/beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
@@ -17,7 +17,8 @@ import Data.Hashable
 import Data.Kind (Type)
 import qualified Data.List.NonEmpty as NE (NonEmpty)
 import Data.Text (Text)
-import Data.Typeable
+import Data.Typeable (Typeable)
+import GHC.Generics (Generic)
 
 -- * Convenience type synonyms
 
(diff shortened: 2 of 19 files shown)


LaurentRDC:
I am really swamped right now so I won't be able to review this for some time. Hopefully someone else can take a look

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
```diff
@@ -184,8 +185,38 @@ class ( IsSql92ColumnConstraintDefinitionSyntax (Sql92ColumnSchemaColumnConstrai
...
+    -- values.
+  | ForeignKeyActionRestrict
+    -- ^ @RESTRICT@: prohibit modification of a row when foreign key references
+    -- to that row exist.
+  deriving (Show, Eq, Ord, Generic)
+instance Hashable ForeignKeyAction
+
 class Typeable constraint => IsSql92TableConstraintSyntax constraint where
-  primaryKeyConstraintSyntax :: [ Text ] -> constraint
+  primaryKeyConstraintSyntax :: [Text] -> constraint
```
LaurentRDC: Can we use `NonEmpty` as much as possible? I keep bringing this up because we have had other issues (e.g. #760) with unexpectedly empty lists

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
```diff
@@ -184,8 +185,38 @@ class ( IsSql92ColumnConstraintDefinitionSyntax (Sql92ColumnSchemaColumnConstrai
                      -> Maybe Text {-^ Default collation -}
                      -> columnSchema
 
+-- | Action to perform on referencing foreign keys when a row is modified.
+data ForeignKeyAction
```
LaurentRDC, line 189: I believe the SQL standard also defines `NO ACTION`, which is admittedly the default, but I think we want to let people choose this specifically if they want to be explicit. 

That way, `Nothing` and `Just ForeignKeyActioNoAction` would behave the same

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
```diff
@@ -184,8 +185,38 @@ class ( IsSql92ColumnConstraintDefinitionSyntax (Sql92ColumnSchemaColumnConstrai
...
+  -- | Emit a @FOREIGN KEY (cols) REFERENCES tbl (refCols)@ table constraint.
+  --
+  -- Use 'addTableForeignKey' to attach a foreign key to a schema definition.
+  foreignKeyConstraintSyntax
+    :: [Text]                 -- ^ local columns
+    -> Text                   -- ^ referenced table name
+    -> [Text]                 -- ^ referenced columns
+    -> Maybe ForeignKeyAction -- ^ ON UPDATE action
+    -> Maybe ForeignKeyAction -- ^ ON DELETE action
+    -> constraint
```
LaurentRDC, line 216: Let's use `NonEmpty` for lists

beam-migrate/Database/Beam/Migrate/Actions.hs
```diff
@@ -279,6 +282,16 @@ createIndexWeight, dropIndexWeight :: Int
...
 
+-- | Witness for whether a backend's table constraint syntax supports
+-- @FOREIGN KEY@ clauses in @CREATE TABLE@ statements.
+data ForeignKeySupport be where
+  -- | The backend supports @FOREIGN KEY@ constraints in @CREATE TABLE@.
+  ForeignKeySupported
+    :: IsSql92ForeignKeyTableConstraintSyntax (BeamSqlBackendTableConstraintSyntax be)
+    => ForeignKeySupport be
+  -- | The backend does not support @FOREIGN KEY@ constraints.
+  ForeignKeyUnsupported :: ForeignKeySupport be
```
LaurentRDC: I'm actually curious if there are any SQL DBMs that do not support foreign keys.
I know some backends (e.g. SQLite) may _ignore_ these constraints, but I wouldn't say they're unsupported.

beam-postgres/test/Database/Beam/Postgres/Test/Migrate.hs
```diff
@@ -192,3 +195,125 @@ uniqueIndexVerification pgConn =
...
+
+instance Table FkParentT where
+  newtype PrimaryKey FkParentT f = FkParentPk (C f Int32)
+    deriving (Generic, Beamable)
+  primaryKey = FkParentPk . _fk_parent_id
+
+data FkChildT f = FkChildT
+  { _fk_child_id        :: C f Int32
+  , _fk_child_parent_id :: C f Int32
+  } deriving (Generic, Beamable)
```
LaurentRDC, line 213: I think this should be

```haskell
data FkChildT f = FkChildT
    { _fk_child_id        :: C f Int32                -- unchanged
    , _fk_child_paremt_id :: PrimaryKey FkParentT f -- new
    } deriving (Generic, Beamable)
```

beam-sqlite/test/Database/Beam/Sqlite/Test/Migrate.hs
```diff
@@ -121,3 +124,120 @@ verifiesUniqueIndex = testCase "verifySchema correctly detects a UNIQUE secondar
...
+
+instance Table ParentT where
+  newtype PrimaryKey ParentT f = ParentPk (C f Int32)
+    deriving (Generic, Beamable)
+  primaryKey = ParentPk . _parent_id
+
+data ChildT f = ChildT
+  { _child_id        :: C f Int32
+  , _child_parent_id :: C f Int32
+  } deriving (Generic, Beamable)
```
LaurentRDC, line 142: Same comment as https://github.com/haskell-beam/beam/pull/791/changes#r3064816214

LaurentRDC (commented):
Thanks for your contribution! This is looking solid.

beam-migrate/Database/Beam/Migrate/SQL/SQL92.hs
sheaf: I don't think we need the `Maybe` if we have an explicit `NO ACTION` constructor, so I removed it.

beam-migrate/Database/Beam/Migrate/Actions.hs
```diff
@@ -377,9 +393,45 @@ createTableActionProvider =
...
+               ForeignKeyUnsupported -> ([], [])
+
+         -- For each foreign key constraint, ensure the referenced table
+         -- already exists in the current state.
+         -- This enforces a topological ordering, so that the solver schedules
+         -- CREATE TABLE for all referenced tables first.
+         --
+         -- Note: this does not support cycles in the foreign key reference graph,
+         -- for which we would need ALTER TABLE ADD CONSTRAINT, which we don't
+         -- currently support.
```
LaurentRDC, line 388: Does the current design *prevent* adding this support in a later patch? What would be needed to get it working?

sheaf: I don't think the current patch would prevent that enhancement. I believe we would need to:

  - add a new typeclass, with an instance for Postgres, 
  - update this logic to allow declaring foreign key constraints that refer to tables that don't exist yet; it would need to branch on whether ALTER TABLE ADD CONSTRAINT is supported, which seems like it would require another GADT like `ForeignKeySupport`
  - add a provider that generates the appropriate ALTER TABLE syntax for foreign key constraints that remain unfulfilled

beam-migrate/Database/Beam/Migrate/Actions.hs
LaurentRDC: @sheaf I think we should drop `ForeignKeyUnsupported`

sheaf: Hmm, I don't completely follow your suggestion. The purpose of the GADT was to allow users to call `createTableActionProvider` without an instance for `IsSql92ForeignKeyTableConstraintSyntax`. If I remove that constructor, then I might as well add foreign key constraint syntax to the `IsSql92TableConstraintSyntax` typeclass, but that seems like a bigger breaking change. Would you be OK with that change?

LaurentRDC: >  I might as well add foreign key constraint syntax to the `IsSql92TableConstraintSyntax` typeclass

Right, I didn't consider this. But at the same time, I'm OK with this breaking change -- I'll just have to version `beam-migrate` and the backends appropriately on the next release, once this is merged

LaurentRDC:
[Just one more comment to address](https://github.com/haskell-beam/beam/pull/791#discussion_r3076268412) and this is good to go!

beam-migrate/Database/Beam/Migrate/Actions.hs
sheaf: OK, I've made that change now. It does seem cleaner that way.

LaurentRDC (approved):

LaurentRDC:
Woohoo! Very cool. I'm going to try and release this change this week

PR #794, opened 2026-04-14
Fixes to beam-migrate solving

sheaf:
I found out that my changes to indices and foreign keys could cause issues with the migration solver:

  - For indices, beam-migrate would fail to make a unique index into a non-unique index, or vice-versa. That's fixed in the first commit.
  - For mismatches in foreign keys constraints, with SQLite we are in a bit of a bind because there is no ALTER TABLE ADD CONSTRAINT functionality available. So we just want the solver to fail to find a migration. It was not doing so because the solver was endlessly trying to drop columns to try to move closer to the goal, even though it was never helpful. So the second commit adds some quick checks to avoid trying migration steps that drop columns that we want in the goal, as this makes the search space grow (which makes the solver basically run forever).

### DIFF
--- a/beam-migrate/ChangeLog.md
+++ b/beam-migrate/ChangeLog.md
@@ -16,6 +16,9 @@
 * Fix an issue in which `beam-migrate` would fail to migrate a unique index
   to a non-unique index or vice-versa.
 
+* Fixed an issue in which the migration solver could end up taking an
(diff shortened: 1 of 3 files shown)


sheaf:
I have now added tests for both of the commits and confirmed that they fail without their associated fixes.

LaurentRDC (approved):

PR #795, opened 2026-04-14
Don't try to migrate SQLite internal tables

sheaf:
The SQLite migration backend incorrectly took into account internal SQLite tables (e.g. sqlite_sequence). It would then try to drop them, making even trivial migrations look like they lose data.

The fix is simple: just filter out SQLite internal tables; they start with `sqlite_` (which is reserved for internal SQLite tables as per the SQLite documentation).


LaurentRDC:
Do you have a minimal test case, that fails without this patch, that we could add to ensure there's no regression here?

sheaf:
I added a test and confirmed that it fails before this patch.

sheaf:
Spurious failure in the 9.12.2 job (needs to be restarted).

LaurentRDC:
Thanks!

PR #796, opened 2026-04-15
Add support for temporary tables

sheaf:
This PR adds support for temporary tables to the SQLite and Postgres backends.

The main logic had to live in `beam-migrate` (not `beam-core`) as that's where the relevant DDL syntax lives (I found this a bit counter-intuitive).

Temporary tables can now be created with `runCreateTempTable`:

  - Creation takes a `TempTableOptions` argument which dictates whether to create the temporary table fresh; see `TempTableCreateMode`.
  - `runCreateTempTable` returns a normal table entity that can be used with other queries like any other table.

### DIFF
--- a/beam-core/ChangeLog.md
+++ b/beam-core/ChangeLog.md
@@ -4,6 +4,13 @@
 
 * Updated the upper bound on `time` to include `time-1.14`
 
+## Query combinator API change
+
+* The `Database be db` context has been removed from the query combinators `all_`,
+  `allFromView_`, `join_`, `join'`, `related_`, `relatedBy_` and `relatedBy_'`.
+
+  This constraint was not used by any of these functions.
+
 # 0.10.4.0
 
 ## Added features
--- a/beam-postgres/test/Database/Beam/Postgres/Test/TempTable.hs
+++ b/beam-postgres/test/Database/Beam/Postgres/Test/TempTable.hs
@@ -1,0 +1,87 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+module Database.Beam.Postgres.Test.TempTable (tests) where
+
+import Data.ByteString (ByteString)
+import Data.Int (Int32)
+import Data.Kind (Type)
+import GHC.Exts (Any)
+
+import Database.Beam
+import Database.Beam.Postgres
+import Database.PostgreSQL.Simple (execute_)
+
+import Test.Tasty
+import Test.Tasty.HUnit
+
+import Database.Beam.Postgres.Test
+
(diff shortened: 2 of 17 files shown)


beam-core/ChangeLog.md
```diff
@@ -4,6 +4,13 @@
 
 * Updated the upper bound on `time` to include `time-1.14`
 
+## Query combinator API change
+
+* The `Database be db` context has been removed from the query combinators `all_`,
+  `allFromView_`, `join_`, `join'`, `related_`, `relatedBy_` and `relatedBy_'`.
+
+  This constraint was not used by any of these functions.
+
```
LaurentRDC, line 13: Interesting!

beam-postgres/test/Database/Beam/Postgres/Test/TempTable.hs
```diff
@@ -0,0 +1,89 @@
...
+import Database.Beam.Postgres.Test
+
+tests :: IO ByteString -> TestTree
+tests getConn = testGroup "Temporary table tests"
+  [ testStageFilteredRows getConn
+  , testDropAndCreate      getConn
+  ]
+
+-- | Phantom db type of the correct kind '(Type -> Type) -> Type'.
+data NoDb (f :: Type -> Type)
```
LaurentRDC: This `NoDb` construct seems like it would be quite useful to provide along with `runCreateTempTable` from `beam-migrate` itself

LaurentRDC (commented):
Interesting feature!

beam-postgres/test/Database/Beam/Postgres/Test/TempTable.hs
sheaf: I'm not sure it's very useful. I removed it and used `Any` instead. Really the point is that it is a phantom parameter so you can change it using `coerce`, to perform queries involving a temporary table against any other database.

LaurentRDC: Since this might not be obvious to others (I didn't know about `Any`!), it would be worth it to either include it in the docstring of `runCreateTempTable` as an example, or put it in the documentation. 
Modifying the documentation remains difficult, so I don't want to require this for this PR.

LaurentRDC:
@sheaf you've been on a roll with incoming PRs -- let me know if you plan to submit others, so I only go through one release cycle containing all the recent changes

sheaf:
> @sheaf you've been on a roll with incoming PRs -- let me know if you plan to submit others, so I only go through one release cycle containing all the recent changes

I don't have anything else upcoming! But no rush, I'm quite happy using `source-repository-package` stanzas until a release.

PR #797, opened 2026-04-15
Performance improvements for beam-postgres

### LINKED ISSUE #769: Postgres de/serialization appears to be 2x slower than postgresql-simple
See discussion here: https://github.com/haskell-beam/beam/discussions/767

LaurentRDC:
This PR is about improving performance of beam-postgres to match postgresql-simple.

Fixes #769

### DIFF
--- a/beam-postgres/ChangeLog.md
+++ b/beam-postgres/ChangeLog.md
@@ -7,6 +7,12 @@
 * `getDbConstraintsForSchemas` now discovers foreign key constraints
   via `pg_constraint`, including `ON DELETE` / `ON UPDATE` actions.
 
+## Performance improvements
+
+* By minimizing redundant work in the hot loop, the performance of `beam-postgres`
+  when fetching data from a database has been improved by 30%. The performance
+  of `beam-postgres` is now within 10% of raw queries via `postgresql-simple` (#797).
+
 # 0.5.5.0
 
 ## Added features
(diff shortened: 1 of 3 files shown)


LaurentRDC:
The baseline is:

```
$ BEAM_BENCH_ROWS=100000 cabal -v0 bench beam-backend-bench:beam-postgres
Starting benchmark with 100000 rows
(set BEAM_BENCH_ROWS to override)
All
  SELECT * FROM imdb_names (100000 rows)
    postgresql-simple: OK
      249  ms ±  12 ms
    beam-postgres:     OK
      321  ms ± 8.8 ms
```

This is showing a 50% slowdown in beam-postgres vs postgresql-simple

LaurentRDC:
After removing some redundant work from parsing every row:

```
$ BEAM_BENCH_ROWS=100000 cabal -v0 bench beam-backend-bench:beam-postgres
Starting benchmark with 100000 rows
(set BEAM_BENCH_ROWS to override)
All
  SELECT * FROM imdb_names (100000 rows)
    postgresql-simple: OK
      243  ms ±  13 ms
    beam-postgres:     OK
      263  ms ± 7.0 ms
```

`beam-postgres` is now showing a ~10% slowdown over `postgresql-simple`

LaurentRDC:
I believe that the remaining slowdown is due to the use of free monads. This means that speeding up the `beam-postgres` benchmark would require a breaking change in `beam-core`, such that backends would need to adjust.

[Thankfully, this shouldn't be too bad, as there is only one "third-party" backend](https://hackage.haskell.org/packages/search?terms=beam) released on Hackage (`beam-mysql`) which would need quite a bit of work to bring back to life anyways.

In any case, this will be left for further work

LaurentRDC:
I am perplexed. The benchmarks on my local machine are showing that this PR closes the gap between `beam-postgres` and `postgresql-simple`. However, Github Actions are NOT showing these gains. I can't explain this

PR #785, opened 2026-03-01
SQLite: flesh out RETURNING clause

sheaf:
This MR continues the work started in 24ecab6 to update how the SQLite backend handles `INSERT ... RETURNING`, in particular extending the logic to support `INSERT ... ON CONFLICT ... RETURNING` as well as `DELETE ... RETURNING`, with support for an additional projection that describes what to return (i.e. filtering the returned value on the database side), instead of always returning everything (`RETURNING *`).

It also updates the general workaround for SQLite handling of `DEFAULT` keywords, to more efficiently group queries by `DEFAULT` column layout.

Tested in `Database.Beam.Sqlite.Test.Returning.`

Issue #775, opened 2025-10-18
MonadBeam...Returning should be generalized


sheaf:
Currently we have:

```hs
class MonadBeam be m => MonadBeamDeleteReturning be m | m -> be where
  runDeleteReturningList
    :: ( Beamable table
       , Projectible be (table (QExpr be ()))
       , FromBackendRow be (table Identity) )
    => SqlDelete be table
    -> m [table Identity]

class MonadBeam be m => MonadBeamInsertReturning be m | m -> be where
  runInsertReturningList
    :: ( Beamable table
       , Projectible be (table (QExpr be ()))
       , FromBackendRow be (table Identity) )
    => SqlInsert be table
    -> m [table Identity]

class MonadBeam be m => MonadBeamUpdateReturning be m | m -> be where
  runUpdateReturningList
    :: ( Beamable table
       , Projectible be (table (QExpr be ()))
       , FromBackendRow be (table Identity) )
    => SqlUpdate be table
    -> m [table Identity]
```

The limitation of these is that they all return the full rows, while one might only want a subset of the rows.

For this reason, the Postgres backend provides the additional functions `runPg...ReturningList` functions, where we see type signatures such as:

```hs
deleteReturning :: Projectible Postgres a
                => DatabaseEntity Postgres be (TableEntity table)
                -> (forall s. table (QExpr Postgres s) -> QExpr Postgres s Bool)
                -> (table (QExpr Postgres PostgresInaccessible) -> a)
                -> PgDeleteReturning (QExprToIdentity a)
```

The key difference is that instead of returning `table Identity`, we pass in a projection function, of type `(table (QExpr Postgres PostgresInaccessible) -> a)`, and we use it to project out the fields we want to return (instead of returning the whole row). So we're replacing `table Identity` with `QExprToIdentity a`, where `a` is the chosen row type to project out.

Given that the SQLite backend supports this feature as well, it seems natural to generalise these monads.

LaurentRDC:
Completed in #801

PR #799, opened 2026-04-21
MonadBeamUpdateReturning Sqlite SqliteM: implement

thomasjm:
Continuing work from https://github.com/haskell-beam/beam/pull/785.

CC @LaurentRDC

### DIFF
--- a/beam-sqlite/Database/Beam/Sqlite/Connection.hs
+++ b/beam-sqlite/Database/Beam/Sqlite/Connection.hs
@@ -24,7 +24,7 @@ module Database.Beam.Sqlite.Connection
 
     -- ** @UPDATE ... RETURNING@
   , SqliteUpdateReturning
-  , updateReturning, runUpdateReturningList
+  , updateReturning, runSqliteUpdateReturningList
   ) where
 
 import           Prelude hiding (fail)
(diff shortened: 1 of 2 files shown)


LaurentRDC:
Awesome!
Can you modify the test suite to use the new method `runUpdateReturningList`?

thomasjm:
Okay, I took the liberty of renaming the `runUpdateReturningList` function from `Database.Beam.Sqlite.Connection` to `runSqliteUpdateReturningList`, so that it wouldn't collide with the one from `MonadBeamUpdateReturning`.

Then I realized this isn't quite as useful as I hoped it was going to be (for unifying my Postgres and Sqlite UPDATE RETURNING code paths) without https://github.com/haskell-beam/beam/issues/775.

So, maybe the breaking function rename I've done here can go into 0.11 along with that.

LaurentRDC:
I think this is perfectly fine. 
Note that this change only touches `beam-sqlite`, which has its own update cadence. So this would be a breaking change of `beam-sqlite` only.

LaurentRDC (approved):

LaurentRDC:
Thanks! I'll make a new release soon

LaurentRDC:
[`beam-sqlite-0.6.0.0`](https://hackage.haskell.org/package/beam-sqlite-0.6.0.0) was just released with your change

PR #800, opened 2026-04-22
MonadBeam*Returning projection support

thomasjm:
This implements the idea from https://github.com/haskell-beam/beam/issues/775 and closes that issue.

I'm happy to re-target this against the `beam-0.11` branch if desired. It might need a rebase on the latest `master` though?

I realized that in #799, I probably should have done `MonadBeamInsertReturning` and `MonadBeamDeleteReturning` as well...

PR #801, opened 2026-04-23
MonadBeam*Returning projection support (against beam-0.11)

### LINKED ISSUE #775: MonadBeam...Returning should be generalized

thomasjm:
(Redux of #800.)

This implements the idea from https://github.com/haskell-beam/beam/issues/775 and closes that issue.

I realized that in #799, I probably should have done `MonadBeamInsertReturning` and `MonadBeamDeleteReturning` as well...

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
+++ b/beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
@@ -41,108 +41,144 @@ import           Data.Semigroup
 
 --import GHC.Generics
 
--- | 'MonadBeam's that support returning the newly created rows of an @INSERT@ statement.
+-- | 'MonadBeam's that support returning data from the newly created rows of an @INSERT@ statement.
 --   Useful for discovering the real value of a defaulted value.
+--
+--   The projection function determines which columns are returned. Pass 'id' to
+--   return the full row ('table Identity'), or supply a custom projection to
+--   return a subset of columns.
 class MonadBeam be m =>
   MonadBeamInsertReturning be m | m -> be where
+  runInsertReturningListWith
+    :: ( Beamable table
+       , Projectible be a
+       , FromBackendRow be (QExprToIdentity a) )
+    => SqlInsert be table
+    -> (table (QExpr be ()) -> a)
+    -> m [QExprToIdentity a]
   runInsertReturningList
     :: ( Beamable table
        , Projectible be (table (QExpr be ()))
        , FromBackendRow be (table Identity) )
     => SqlInsert be table
     -> m [table Identity]
+  runInsertReturningList sql = runInsertReturningListWith sql id
 
 instance MonadBeamInsertReturning be m => MonadBeamInsertReturning be (ExceptT e m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 instance MonadBeamInsertReturning be m => MonadBeamInsertReturning be (ContT r m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 instance MonadBeamInsertReturning be m => MonadBeamInsertReturning be (ReaderT r m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 instance MonadBeamInsertReturning be m => MonadBeamInsertReturning be (Lazy.StateT r m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 instance MonadBeamInsertReturning be m => MonadBeamInsertReturning be (Strict.StateT r m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 instance (MonadBeamInsertReturning be m, Monoid r)
     => MonadBeamInsertReturning be (Lazy.WriterT r m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 instance (MonadBeamInsertReturning be m, Monoid r)
     => MonadBeamInsertReturning be (Strict.WriterT r m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 instance (MonadBeamInsertReturning be m, Monoid w)
     => MonadBeamInsertReturning be (Lazy.RWST r w s m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 instance (MonadBeamInsertReturning be m, Monoid w)
     => MonadBeamInsertReturning be (Strict.RWST r w s m) where
-    runInsertReturningList = lift . runInsertReturningList
+    runInsertReturningListWith sql proj = lift $ runInsertReturningListWith sql proj
 
--- | 'MonadBeam's that support returning the updated rows of an @UPDATE@ statement.
+-- | 'MonadBeam's that support returning data from the updated rows of an @UPDATE@ statement.
 --   Useful for discovering the new values of the updated rows.
+--
+--   The projection function determines which columns are returned. Pass 'id' to
+--   return the full row ('table Identity'), or supply a custom projection to
+--   return a subset of columns.
 class MonadBeam be m =>
   MonadBeamUpdateReturning be m | m -> be where
+  runUpdateReturningListWith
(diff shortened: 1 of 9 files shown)


LaurentRDC:
This is looking pretty good!
Would you be able to add an item in the changelog for `beam-core` related to this change?

beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
```diff
@@ -41,108 +41,123 @@ import           Data.Semigroup
...
   MonadBeamInsertReturning be m | m -> be where
   runInsertReturningList
     :: ( Beamable table
-       , Projectible be (table (QExpr be ()))
-       , FromBackendRow be (table Identity) )
+       , Projectible be a
+       , FromBackendRow be (QExprToIdentity a) )
     => SqlInsert be table
-    -> m [table Identity]
+    -> (table (QExpr be ()) -> a)
```
LaurentRDC, line 57: When looking at the examples (e.g. in the test cases), it seems to me that it's quite annoying to have the projection _after_ the Insert / Update / Delete. E.g.:

```haskell
runInsertReturningList 
    (
    <big expression>
    ) 
    id
```

Instead, I think the arguments should be flipped:

```haskell
runInsertReturningList id $
    <big expression>
```

The same comment applies to `runUpdateReturningList` and `runDeleteReturningList`

beam-sqlite/test/Database/Beam/Sqlite/Test/Insert.hs
```diff
@@ -56,11 +56,12 @@ testInsertReturningColumnOrder = testCase "runInsertReturningList with mismatchi
...
-      insert (dbTestTable testDatabase) $
+      runBeamSqlite conn $ runInsertReturningList
+      (insert (dbTestTable testDatabase) $
       insertExpressions [ TestTable 0 (concat_ [ "j", "im" ]) "smith" 19 currentTimestamp_ (val_ zeroUtcTime)
                         , TestTable 1 "sally" "apple" ((val_ 56 + val_ 109) `div_` 5) currentTimestamp_ (val_ oneUtcTime)
-                        , TestTable 4 "blah" "blah" (-1) currentTimestamp_ (val_ now) ]
+                        , TestTable 4 "blah" "blah" (-1) currentTimestamp_ (val_ now) ])
+      id
 
     let dateJoined : _ = ttDateJoined <$> inserted
```
LaurentRDC, line 65: Here's a good opportunity to use `ttDateJoined` as a projection, rather than `id` on line 64

beam-postgres/test/Database/Beam/Postgres/Test/Marshal.hs
```diff
@@ -173,7 +173,7 @@ marshalTest' cmp gen postgresConn =
 
       [MarshalTable rowId v] <-
         liftIO . runBeamPostgres conn $
-        runInsertReturningList $ insert (_marshalTbl marshalDb) $ insertExpressions [ MarshalTable default_ (val_ a) ]
+        runInsertReturningList (insert (_marshalTbl marshalDb) $ insertExpressions [ MarshalTable default_ (val_ a) ]) id
```
LaurentRDC: Here's another good opportunity to replace `id` with a projection for the second column of `MarshalTable`

beam-sqlite/test/Database/Beam/Sqlite/Test/Insert.hs
thomasjm: The full rows are used just below this, we'd have to remove the assertEqual on them.

beam-postgres/test/Database/Beam/Postgres/Test/Marshal.hs
thomasjm: Similar problem here -- you're saying we can have the projection be `_marshalTableEntry`, so we'd just get back `v`? But the test also makes use of the `rowId` here. Please let me know if I'm misunderstanding.

beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
thomasjm: I made this change in the code, but then I looked at it and realized that it looked un-Beam-like to me and had to think about why.

Most beam functions accept arguments in a logical order that corresponds to the SQL query. In this case the query is "UPDATE ... SET ... WHERE ... RETURNING ...". The projection corresponds to the RETURNING part, so that's why it goes last. Existing code like the [Postgres updateReturning](https://hackage-content.haskell.org/package/beam-postgres-0.5.5.0/docs/src/Database.Beam.Postgres.Full.html#updateReturning) are written with this convention so I think it's probably best to stick with it.

thomasjm:
Thanks for the review! I added the changelog entries, but had some follow-up questions about the other comments.

beam-sqlite/test/Database/Beam/Sqlite/Test/Insert.hs
LaurentRDC: Ahh, understood

beam-postgres/test/Database/Beam/Postgres/Test/Marshal.hs
LaurentRDC: I apologize, I didn't check that `rowId` was used below. The standard PR view only shows the like where `v` is checked. Carry on!

beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
LaurentRDC: That's fair, let's keep it as it was initially

beam-core/Database/Beam/Backend/SQL/BeamExtensions.hs
```diff
@@ -41,108 +41,123 @@ import           Data.Semigroup
...
--- | 'MonadBeam's that support returning the newly created rows of an @INSERT@ statement.
+-- | 'MonadBeam's that support returning data from the newly created rows of an @INSERT@ statement.
 --   Useful for discovering the real value of a defaulted value.
+--
+--   The projection function determines which columns are returned. Pass 'id' to
+--   return the full row ('table Identity'), or supply a custom projection to
+--   return a subset of columns.
 class MonadBeam be m =>
   MonadBeamInsertReturning be m | m -> be where
   runInsertReturningList
```
LaurentRDC: I had another idea. Right now, the signature for `runXReturningList` is changing, which is going to cause churn in downstream users.

What if we instead changed the method to be called e.g. `runInsertReturningListWith`, which can take a projection, and then have a "helper function" `runInsertReturningList = flip runInsertReturningListWith id`.

That way, people can keep using `runInsertReturningList` as they have previously, without breakage. The same comment would apply to `runUpdateReturningList` and `runDeleteReturningList`.

What do you think?

thomasjm: Sounds good to me, done in latest commit.

LaurentRDC (approved):

LaurentRDC:
Thank you for your contribution!

thomasjm:
Thank you! No rush, but do you have any idea when 0.11 might be released?

LaurentRDC:
> Thank you! No rush, but do you have any idea when 0.11 might be released?

It's been in preparation for years at this point ... It's time. Let me polish it up and release it in the next week

LaurentRDC:
@thomasjm `beam-core-0.11` has been released (as well as all other `beam-*` packages to support it)

thomasjm:
Awesome, thanks @LaurentRDC !

PR #803, opened 2026-04-23
Fix stale fieldsCache in withPgDebug across AtOnce queries

### LINKED ISSUE #802: fieldsCache in withPgDebug causes stale field metadata across queries (regression from #797)
## Bug

PR #797 introduced a one-entry `fieldsCache` in `withPgDebug` to avoid redundant `getFields` / `nfields` / `ftype` calls within a cursor batch. The cache is an `IORef (Maybe (Vector Pg.Field))` with no key:

```haskell
case cached of
  Just fs -> pure fs   -- returns stale fields regardless of which Pg.Result produced them
```

When two queries with different column types are issued within the same `runBeamPostgres` call, the second query reads field OID metadata from the first. `postgresql-simple`'s `fromField` rejects the wrong OID and throws, producing:

```
BeamRowReadError {brreColumn = Just 0, brreError = ColumnErrorInternal "Column parse failed with unknown exception"}
```

The existing comment at the call site already describes the intended behavior ("Caching by `Pg.Result` equality") — the implementation just didn't do it.

## Reproduction

```haskell
execute_ conn "CREATE TABLE alpha (name TEXT  NOT NULL)"
execute_ conn "CREATE TABLE beta  (value INT4 NOT NULL)"
execute_ conn "INSERT INTO alpha VALUES ('hello')"
execute_ conn "INSERT INTO beta  VALUES (42)"

-- Fails with BeamRowReadError on the second query:
(names, values) <- runBeamPostgres conn $ do
  ns <- runSelectReturningList $ select $ all_ (dbAlpha db)
  vs <- runSelectReturningList $ select $ all_ (dbBeta  db)
  pure (ns, vs)
```

## Root cause

`renderExecReturningList` (the `AtOnce` path used by `runReturningList`) calls `cachedGetFields res` and then frees `res` via `Pg.unsafeFreeResult` after the rows are consumed. libpq may reuse the freed pointer address for the next `Pg.exec`, meaning the cache key comparison `cachedRes == res` yields a false hit on the next query even with a keyed cache.

The cursor-batch path (`stepProcess`) does not call `unsafeFreeResult`, so pointer-keyed caching is safe there.

## Fix

`renderExecReturningList` calls `getFields res` directly (no cache). The cache remains in place for `stepProcess` where it is both safe and beneficial (avoids repeated `getFields` calls across rows within a 256-row batch).

compilerbauer:
## Summary

- `renderExecReturningList` (the `AtOnce` path) was calling `cachedGetFields` and then freeing the result via `Pg.unsafeFreeResult`. libpq can reuse the freed pointer address on the next `Pg.exec`, causing the next query to get a false cache hit and inherit stale field OID metadata from the previous query.
- Fix: `renderExecReturningList` now calls `getFields` directly. The `cachedGetFields` one-entry cache is preserved for `stepProcess` (CursorBatching), where `unsafeFreeResult` is never called and the cache is both safe and beneficial (avoids `getFields` per row within a 256-row batch).
- Regression test added: two tables with different column types (`TEXT` vs `INT4`) queried in sequence within one `runBeamPostgres` call.

Fixes #802. Regression introduced in #797.

## Test plan

- [ ] `cabal test beam-postgres-tests --test-options="--pattern fieldsCache"` passes
- [ ] Full `cabal test beam-postgres-tests` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Connection.hs
+++ b/beam-postgres/Database/Beam/Postgres/Connection.hs
@@ -197,16 +197,16 @@ withPgDebug dbg conn (Pg action) = do
   -- Default batch size is 256 rows, set by 'postgresql-simple'
   -- 'defaultFoldOptions' (FetchQuantity = Automatic, which resolves to
   -- 256 in 'Database.PostgreSQL.Simple').
-  fieldsCache <- newIORef (Nothing :: Maybe (Vector Pg.Field))
+  fieldsCache <- newIORef (Nothing :: Maybe (Pg.Result, Vector Pg.Field))
 
   let cachedGetFields :: Pg.Result -> IO (Vector Pg.Field)
       cachedGetFields res = do
         cached <- readIORef fieldsCache
         case cached of
-          Just fs -> pure fs
+          Just (cachedRes, fs) | cachedRes == res -> pure fs
           _ -> do
             fs <- getFields res
-            writeIORef fieldsCache (Just fs)
+            writeIORef fieldsCache (Just (res, fs))
             pure fs
 
       finish x = pure (Right x)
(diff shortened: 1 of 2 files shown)


LaurentRDC:
Ooof, thanks for sending a patch. This is very unfortunate

LaurentRDC:
[beam-postgres-0.5.6.0 has been deprecated in Hackage](https://hackage.haskell.org/package/beam-postgres-0.5.6.0). I'm preparing release 0.5.6.1 which will include this fix. It will be released tomorrow

compilerbauer:
Thank you!

LaurentRDC:
[`beam-postgres-0.5.6.1`](https://hackage.haskell.org/package/beam-postgres-0.5.6.1) has been released with this patch. Thank you!

PR #804, opened 2026-04-27
Documentation fixes

### LINKED ISSUE #580: Problems in tutorial 3
I recently reproduced all 3 of the beam tutorials in a slightly different context, and while doing so, I found some strange results. I'll record them below; it's not totally obvious to me if these are a result of a problem in the tutorial or a problem in beam itself (or indeed a misunderstanding on my part).

## `allUsersAndTotals`

The tutorial gives two versions for the query for `allUsersAndTotals`. The text mentions that the second one could perform significantly better than the first; from this I gather that both queries are intended to be equivalent. However, the first query returns

```
(betty,Just 16500)
(james,Just 0)
(sam,Just 0)
```

which is incorrect, as it has `james` total at zero. The second query does give the correct result:

```
(james,Just 27000)
(betty,Just 16500)
(sam,Just 0)
```

## `shippingInformationByUser`

For this query the tutorial gives _three_ variants. For the first query, the tutorial correctly points out that it computes an incorrect result:

```
(betty,0,1)
(james,2,0)
(sam,1,0)
```

This is incorrect, because `sam` placed no orders at all, and so can't have any unshipped orders either. The tutorial then goes on to fix the query, _but_ the result of that new fixed query is

```
(betty,1,1)
(james,2,1)
(sam,1,1)
```

Now the results are off for _all_ users! Finally, the third variant of the query (with explicit `subselect_`) also returns this same, incorrect (I think), result.

## `NoMonomorphismRestriction`

Finally, a minor comment. In tutorial2, it is important that in 

```haskell
let james = User "james@example.com" "James" "Smith" "b4cc344d25a2efe540adbf2678e2304c"
    betty = User "betty@example.com" "Betty" "Jones" "82b054bd83ffad9b6cf8bdb98ce3cc2f"
    sam = User "sam@example.com" "Sam" "Taylor" "332532dcfaa1cbf61e2a266bd723612c"
```

a polymorphic type is inferred for these values (that is, `forall f. IsString (Columnar f Text) => UserT f`): this is important because these values are sometimes used in the tutorial at `UserT Identity` (for example, in `insertValues`), and sometimes at `UserT (QExpr Sqlite s)`, for example when using `pk james` etc. in the definition of the `addresses` (rather than `val_ (pk james)`). This requires the `NoMonomorphismRestriction` to be enabled. Might be worth pointing that out to avoid some confusion.

LaurentRDC:
This pull request fixes a few documentation issues, which can now be represented on the website. Each commit is linked to an issue identified in #580.

Fixes #580

### DIFF
--- a/docs/tutorials/tutorial2.md
+++ b/docs/tutorials/tutorial2.md
@@ -246,6 +246,18 @@ runBeamSqliteDebug putStrLn conn $ runInsert $
   insertValues [ james, betty, sam ]
 ```
 
+!!! tip "Tip"
+    The `let` bindings for `james`, `betty`, and `sam` above are reused at two
+    different types: as concrete rows (`UserT Identity`) inside
+    `insertValues`, and as expression-level values (here as `pk james`, of
+    type `PrimaryKey UserT (QExpr Sqlite s)`) inside `insertExpressions`.
+    For the same binding to be usable at both types, GHC must infer a
+    polymorphic type for it. Inside GHCi this happens by default, but in a
+    compiled module you may need to enable `NoMonomorphismRestriction` (or
+    give each binding an explicit polymorphic signature). Otherwise the
+    monomorphism restriction will pin `james` to whichever type it is first
+    used at, and the second use will fail to typecheck.
+
 Now that we have some `User` objects, we can create associated addresses. Notice
 that above, we used `insertValues` to insert concrete `User` rows. This worked
 because we could determine every field of `User` before insertion. `Address`es
--- a/docs/tutorials/tutorial3.md
+++ b/docs/tutorials/tutorial3.md
@@ -483,7 +483,25 @@ Some RDBMSs, like Postgres, given such a query will be unable to utilize availab
 to perform join operations - this translates to *extremely* poor perfomance for even moderately
 sized data.
 
+!!! warning "Nullable columns and `maybe_` on a left-joined table"
+    There is also a *correctness* gotcha here whenever the left-joined table
+    contains a schema-level nullable column. `maybe_` (and `isJust_`) on a
+    nullable table is implemented as "every column of the row is non-`NULL`",
+    so a real, matched row whose nullable column happens to be `NULL` is
+    treated as `Nothing`. In our schema `OrderT` has the nullable
+    `_orderShippingInfo` column, so for any of James's unshipped orders
+    `maybe_ (val_ False) ... order` evaluates to `False`, and the chained
+    `lineItem` and `product` left joins return no rows for him. The query
+    above will compute `Just 0` for James instead of `Just 27000`.
+
+    The next variant uses `leftJoin_'` and avoids `maybe_` over a nullable
+    table entirely, so it is both faster *and* correct in this case. As a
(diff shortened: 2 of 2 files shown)


LaurentRDC:
I'm requesting some extra eyes on this because it's not clear to me if the implementation of `isJust_` was designed this way.

The original generated SQL query is:

```sql
SELECT 
  "t0"."email" AS "res0", 
  "t0"."first_name" AS "res1", 
  "t0"."last_name" AS "res2", 
  "t0"."password" AS "res3", 
  SUM(
    (
      CASE WHEN ("t2"."item_quantity") IS NOT NULL THEN "t2"."item_quantity" ELSE ? END
    ) * (
      CASE WHEN ("t3"."price") IS NOT NULL THEN "t3"."price" ELSE ? END
    )
  ) AS "res4" 
FROM 
  "cart_users" AS "t0" 
  LEFT JOIN "orders" AS "t1" ON ("t1"."for_user__email")=("t0"."email") 
  LEFT JOIN "line_items" AS "t2" ON CASE WHEN (
    (
      (
        (
          ("t1"."id") IS NOT NULL
        ) 
        AND (
          ("t1"."date") IS NOT NULL
        )
      ) 
      AND (
        ("t1"."for_user__email") IS NOT NULL
      )
    ) 
    AND (
      ("t1"."ship_to_address__id") IS NOT NULL
    )
  ) 
  AND (
    ("t1"."shipping_info__id") IS NOT NULL
  ) THEN ("t2"."item_in_order__id")=("t1"."id") ELSE ? END 
  LEFT JOIN "products" AS "t3" ON CASE WHEN (
    (
      ("t2"."item_in_order__id") IS NOT NULL
    ) 
    AND (
      ("t2"."item_for_product__id") IS NOT NULL
    )
  ) 
  AND (
    ("t2"."item_quantity") IS NOT NULL
  ) THEN ("t2"."item_for_product__id")=("t3"."id") ELSE ? END 
GROUP BY 
  "t0"."email", 
  "t0"."first_name", 
  "t0"."last_name", 
  "t0"."password" 
ORDER BY 
  SUM(
    (
      CASE WHEN ("t2"."item_quantity") IS NOT NULL THEN "t2"."item_quantity" ELSE ? END
    ) * (
      CASE WHEN ("t3"."price") IS NOT NULL THEN "t3"."price" ELSE ? END
    )
  ) DESC;
```

whereas with the third commit here, it becomes:

```sql
...
  LEFT JOIN "line_items" AS "t2" ON CASE WHEN NOT(
    (
      (
        (
          (
            ("t1"."id") IS NULL
          ) 
          AND (
            ("t1"."date") IS NULL
          )
        ) 
        AND (
          ("t1"."for_user__email") IS NULL
        )
      ) 
      AND (
        ("t1"."ship_to_address__id") IS NULL
      )
    ) 
    AND (
      ("t1"."shipping_info__id") IS NULL
    )
  ) THEN ("t2"."item_in_order__id")=("t1"."id") ELSE ? END
...
```

LaurentRDC:
Hmm this is basically the same "fix" as https://github.com/haskell-beam/beam/pull/442.
I'll make a documentation edit to explain why the second query should be used instead of the first one.

PR #806, opened 2026-05-06
Add support for COPY statements

### LINKED ISSUE #437: Where's postgres COPY?
Hi,

According to the [`beam-core` haddocks,](http://hackage.haskell.org/package/beam-core-0.8.0.0/docs/Database-Beam.html)

> More complicated strategies (for example, Postgres's COPY) are supported in individual backends.

I was unable to find this in the `beam-postgres` docs. I've tried google, and I've looked for `copy` and `runCopy` in the index.

Is this implemented? If so, could the docs be updated with a link?

Thank you.

### LINKED ISSUE #782: Add support for `COPY` statement
[DuckDB supports exporting data using the `COPY` statement](https://duckdb.org/docs/stable/sql/statements/copy). We should support this! This could be very useful for e.g. data conversion from CSV to Parquet.

Note that since Postgres also supports this (and maybe SQLite too!), we might want to create a class for backends to implement this command

LaurentRDC:
This PR adds support for `COPY ... TO` and `COPY ... FROM` in `beam-core`, as a SQL extension (like `INSERT ... RETURNING`).

Features:

* File-based export with `copyTableTo` and `copySelectTo`;
* File based import with `copyTableFrom`;
* Stream-based export with `copyTableToStream` and `copySelectToStream`;
* Stream-based import with `copyTableFromStream`.

The Postgres backend support both file-based and stream-based copy, while the DuckDB backend only support file-based copy for now -- extra support is needed in `duckdb-simple` for stream copy.

Fixes #437, #782


beam-duckdb/tests/Database/Beam/DuckDB/Test/Extensions/DataSource.hs
LaurentRDC, line 1: Just a reshuffling

PR #812, opened 2026-05-13
Support for the migration framework in beam-duckdb

LaurentRDC:
This pull request exposes a new module, `Database.Beam.DuckDB.Migrate`, which allows the migration framework from `beam-migrate` to work with `beam-duckdb`.


LaurentRDC:
One more thing before I merge: I should implement a test that rolls multiple migrations
Edit: done ✅

beam-duckdb/examples/ExamScores.hs
LaurentRDC, line 1: This is just formatting with ormolu

PR #814, opened 2026-05-14
Fix runtime SQL error for `selectWith` without CTEs

### LINKED ISSUE #760: `selectWith . pure` generates invalid SQL

LaurentRDC:
This is the "quick fix" for #760. Using `selectWith` / `pgSelectWith` without a CTE will result in a generated `SELECT`, rather than an empty `WITH` clause.

Ideally, we would also update `withSyntax` and `withRecursiveSyntax` to have `NonEmpty` as input (instead of `[]`), but this is a breaking change for version 0.12.

This PR partially supercedes #761 

Fixes #760

### DIFF
--- a/beam-duckdb/tests/Database/Beam/DuckDB/Test/Query.hs
+++ b/beam-duckdb/tests/Database/Beam/DuckDB/Test/Query.hs
@@ -12,6 +12,7 @@
 module Database.Beam.DuckDB.Test.Query (tests) where
 
 import Control.Monad (void)
+import Data.Functor ((<&>))
 import Data.Int (Int32)
 import Data.List (nub, nubBy, sort, sortOn)
 import Data.Text (Text)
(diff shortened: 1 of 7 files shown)


beam-duckdb/tests/Database/Beam/DuckDB/Test/Query.hs
```diff
@@ -421,6 +423,29 @@ testCommonTableExpression = testCase "Non-recursive common table expression" $ d
 
       rows @?= [("Alice", 3), ("Bob", 1)]
 
+-- Regression test for https://github.com/haskell-beam/beam/issues/760
```
LaurentRDC, line 426: The easiest way to test a `beam-core` change at this time

PR #816, opened 2026-06-06
Cookbook recipes

### LINKED ISSUE #810: Scaffold cookbook recipes
I would like to replicate [Servant's cookbook recipes](https://docs.servant.dev/en/0.20.3.0/cookbook/index.html) to help people writing Beam-based software. 

The idea would be to have SQL queries and their matching Beam implementation. [This is how Rel8 does it](https://rel8.readthedocs.io/en/latest/cookbook.html), for example.

This issue should be considered completed when there's a well-defined process to include recipes in the cookbook, such that people can easily contribute recipes.

LaurentRDC:
This is an initial implementation of a cookbook.

Fixes #810


LaurentRDC:
My nix setup is broken so it's hard to test locally ...

PR #818, opened 2026-06-07
Fixes #448, CI Text instance

### LINKED ISSUE #448: `CI Text` should be usable with `to_tsvector`
In postgresql, the following works fine:

```
CREATE EXTENSION citext;
CREATE TABLE examples ( example citext PRIMARY KEY);
INSERT INTO examples(example) VALUES ('my cool example') ('my even cooler example') ('my lame example');
SELECT * FROM examples WHERE to_tsvector(examples.example) @@ to_tsquery('cool');
```

However, in Beam, you can't use [`toTsvector`](https://hackage.haskell.org/package/beam-postgres-0.3.2.2/docs/Database-Beam-Postgres.html#v:toTsVector) on columns of type `CI Text`.

I see two (relatively) easy way of solving this:

1. Add an instance of [`IsSqlExpressionStringType`](https://hackage.haskell.org/package/beam-core-0.7.2.2/docs/Database-Beam-Backend-SQL-SQL92.html#t:IsSqlExpressionSyntaxStringType) for `CI Text` 
2. Create a new class, named something like `IsValidTsquerySource`, add `Text` and `CI Text` to it, and then use that as the constraint for `toTsvector` instead 

I can implement either (I think), but wanted to ask before writing any code.

blamario:
Fixes #448

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Types.hs
+++ b/beam-postgres/Database/Beam/Postgres/Types.hs
@@ -185,6 +185,9 @@ type instance BeamSqlBackendCopyFromStreamSyntax Postgres = PgCopyFromStreamSynt
 instance BeamSqlBackendIsString Postgres String
 instance BeamSqlBackendIsString Postgres Text
 
+instance BeamSqlBackendIsString Postgres (CI String)
+instance BeamSqlBackendIsString Postgres (CI Text)
+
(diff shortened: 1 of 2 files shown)


LaurentRDC:
Thanks!

PR #821, opened 2026-07-11
Add support for Data Modifying Common Table Expressions (CTEs) in PostgreSQL backend

### LINKED ISSUE #290: Support Data-modifying CTE's
This is a feature I'd love to have, as I use it all the time to write clear, safe insert transactions for junction tables. Since junction tables show up fairly frequently, I think this feature would be a huge win. SO question showing an example:

https://stackoverflow.com/questions/20561254/insert-data-in-3-tables-at-a-time-using-postgres

This would just have to be an added `RETURNING` statement to the upcoming `WITH` support. As far as I know only Postgres supports it (Oracle maybe, not sure). Here are the actual docs on it

https://www.postgresql.org/docs/current/static/queries-with.html#QUERIES-WITH-MODIFYING

kushagarr:
## Summary

This PR adds PostgreSQL-specific support for data-modifying and explicitly
materialized common table expressions in `beam-postgres`, without changing
the portable CTE API or SQL99 abstractions in `beam-core`.

The implementation introduces a placement-indexed `PgWith` builder which
captures PostgreSQL’s rule that data-modifying CTEs must be attached to a
top-level statement. It supports `WITH` statements terminating in `SELECT`,
`INSERT`, `UPDATE`, or `DELETE`, while preserving compatibility with existing
portable `With Postgres` helpers.

A complete `WITH ... <terminal statement>` is represented as one Beam
statement and sent to PostgreSQL in one database round trip.

## PostgreSQL-specific builder

`PgWith` is indexed by `PgCtePlacement`:

- `PgCteNestedAllowed` represents SELECT-only CTE blocks which may be embedded
  in a subquery.
- `PgCteTopLevelOnly` represents blocks containing data-modifying CTEs.

`pgSelectWithNested` accepts only nested-safe blocks.
`pgSelectWithTopLevel`, `pgInsertWith`, `pgUpdateWith`, and `pgDeleteWith`
produce top-level statements and therefore accept either placement.

The placement parameter has a nominal role, preventing `Data.Coerce` from
relabelling a top-level-only block as nested-safe.

Existing helpers built with `With Postgres` can be composed using `pgLiftWith`.
Lifted and PostgreSQL-specific CTEs use the same name supply, so their generated
CTE names cannot collide.

Recursive construction is restricted to SELECT-only blocks. `pgToTopLevel`
allows a completed recursive SELECT block to feed a later data-modifying CTE
without permitting a modifying CTE to refer recursively to itself.

## Supported CTEs

SELECT CTEs can be introduced with:

- `pgSelecting`
- `pgSelectingWith PgCteDefault`
- `pgSelectingWith PgCteMaterialized`
- `pgSelectingWith PgCteNotMaterialized`

Explicit `MATERIALIZED` and `NOT MATERIALIZED` modifiers require PostgreSQL 12
or later. `PgCteDefault` emits no modifier and retains PostgreSQL’s normal
planner behaviour and compatibility with older server versions.

Reusable data-modifying CTEs are provided by:

- `cteInsertReturning`
- `cteUpdateReturning`
- `cteDeleteReturning`

Their `RETURNING` output is exposed through `ReusableQ` and can be consumed by
later CTEs or the terminal statement using `reuse`.

Side-effect-only variants are also provided:

- `cteInsert`
- `cteUpdate`
- `cteDelete`

These omit `RETURNING`, return `()`, and still execute when the surrounding
top-level statement runs.

## Zero-column projections

Reusable CTE projections containing no fields are supported.

For a zero-column SELECT CTE, Beam omits PostgreSQL’s optional CTE column-alias
list. This preserves the query’s row cardinality without exposing any result
fields.

PostgreSQL requires a data-modifying `RETURNING` clause to contain at least one
expression. For a zero-field `cteInsertReturning`, `cteUpdateReturning`, or
`cteDeleteReturning` projection, Beam therefore emits a private
`NULL::boolean` sentinel:

```sql
WITH "cte0"("res0") AS
       (DELETE FROM "source"
        WHERE ...
        RETURNING NULL::boolean)
SELECT FROM "cte0" AS "t0"
```

The sentinel is not exposed to Beam’s result decoder. It preserves one
zero-field result row for every affected row, allowing the reusable result to
participate in `COUNT(*)`, `EXISTS`, joins, and repeated reuse.

When no later CTE or terminal statement needs the affected rows, the
side-effect-only builders should be used instead; they omit `RETURNING`
entirely.

## Terminal statements and no-op handling

A `PgWith` block may terminate with:

- `pgSelectWithTopLevel`
- `pgInsertWith`
- `pgUpdateWith`
- `pgDeleteWith`

The DML consumers retain Beam’s existing `SqlInsert`, `SqlUpdate`, and
`SqlDelete` wrappers, so the existing PostgreSQL `returning` API remains
available for the terminal statement.

Empty inserts and identity updates remain Beam no-ops. Since PostgreSQL cannot
execute a bare `WITH` block, accumulated CTE definitions are discarded when
there is no terminal statement.

This PR also fixes `pgSelectWith` with an empty CTE block: it now renders the
underlying SELECT directly instead of producing an invalid empty `WITH`
clause.

## Documentation

The Haddock documentation and PostgreSQL user guide include paired Beam and
generated-SQL examples covering:

- lifted portable helpers
- explicit materialization
- reusable and side-effect-only modifying CTEs
- zero-column projections
- SELECT, INSERT, UPDATE, and DELETE terminal statements

The Chinook documentation fixture is pinned to the revision matching its
verified checksum, making clean documentation builds reproducible.

## Testing

The test suite covers:

- mixed SELECT, INSERT, UPDATE, and DELETE CTE definitions
- SELECT and DML terminal consumers
- `MATERIALIZED` and `NOT MATERIALIZED`
- reusable and side-effect-only modifying CTEs
- recursive SELECT CTEs feeding later modifications
- parameter ordering across CTE bodies and terminal statements
- shared naming across lifted and native builders
- empty inserts and identity updates
- zero-column SELECT and data-modifying projections
- zero-column cardinality through `COUNT(*)`, `EXISTS`, joins, and repeated reuse
- negative type tests for invalid nesting, recursive modification, coercion,
  and reuse of side-effect-only CTE results
- Hedgehog properties comparing PostgreSQL execution with pure Haskell models

### DIFF
--- a/beam-core/Database/Beam/Backend/SQL/SQL99.hs
+++ b/beam-core/Database/Beam/Backend/SQL/SQL99.hs
@@ -9,6 +9,7 @@ module Database.Beam.Backend.SQL.SQL99
   , IsSql99AggregationExpressionSyntax(..)
   , IsSql99CommonTableExpressionSelectSyntax(..)
   , IsSql99CommonTableExpressionSyntax(..)
+  , IsSql99DataModifyingCommonTableExpressionSyntax(..)
   , IsSql99RecursiveCommonTableExpressionSelectSyntax(..)
   , IsSql99SelectSyntax(..)
   , IsSql99DataTypeSyntax(..) ) where
@@ -67,3 +68,19 @@ class IsSql99CommonTableExpressionSyntax syntax where
   type Sql99CTESelectSyntax syntax :: Type
 
   cteSubquerySyntax :: Text -> [Text] -> Sql99CTESelectSyntax syntax -> syntax
+
+-- | Extension of SQL99 common-table-expression syntax for backends that allow
+-- a CTE body to be a data-modifying statement rather than a @SELECT@.
+--
+-- The data-modifying body is kept distinct from 'Sql99CTESelectSyntax' so a
+-- backend must opt into this extension explicitly. 'cteDataModifyingSyntax'
+-- supplies the CTE name and output column names around a backend-specific
+-- statement such as @DELETE ... RETURNING ...@.
+class IsSql99CommonTableExpressionSyntax syntax
+    => IsSql99DataModifyingCommonTableExpressionSyntax syntax where
+
+  -- | Backend-specific syntax for the statement inside the CTE body.
+  type Sql99CTEDataModifyingSyntax syntax :: Type
+
+  -- | Wrap a data-modifying statement as one named CTE definition.
+  cteDataModifyingSyntax :: Text -> [Text] -> Sql99CTEDataModifyingSyntax syntax -> syntax
--- a/beam-core/Database/Beam/Backend/SQL/SQL99.hs
+++ b/beam-core/Database/Beam/Backend/SQL/SQL99.hs
@@ -9,7 +9,6 @@ module Database.Beam.Backend.SQL.SQL99
   , IsSql99AggregationExpressionSyntax(..)
   , IsSql99CommonTableExpressionSelectSyntax(..)
   , IsSql99CommonTableExpressionSyntax(..)
-  , IsSql99DataModifyingCommonTableExpressionSyntax(..)
   , IsSql99RecursiveCommonTableExpressionSelectSyntax(..)
   , IsSql99SelectSyntax(..)
   , IsSql99DataTypeSyntax(..) ) where
@@ -68,19 +67,3 @@ class IsSql99CommonTableExpressionSyntax syntax where
   type Sql99CTESelectSyntax syntax :: Type
 
   cteSubquerySyntax :: Text -> [Text] -> Sql99CTESelectSyntax syntax -> syntax
-
--- | Extension of SQL99 common-table-expression syntax for backends that allow
--- a CTE body to be a data-modifying statement rather than a @SELECT@.
---
--- The data-modifying body is kept distinct from 'Sql99CTESelectSyntax' so a
--- backend must opt into this extension explicitly. 'cteDataModifyingSyntax'
--- supplies the CTE name and output column names around a backend-specific
--- statement such as @DELETE ... RETURNING ...@.
-class IsSql99CommonTableExpressionSyntax syntax
-    => IsSql99DataModifyingCommonTableExpressionSyntax syntax where
-
-  -- | Backend-specific syntax for the statement inside the CTE body.
-  type Sql99CTEDataModifyingSyntax syntax :: Type
-
-  -- | Wrap a data-modifying statement as one named CTE definition.
-  cteDataModifyingSyntax :: Text -> [Text] -> Sql99CTEDataModifyingSyntax syntax -> syntax
--- a/beam-postgres/Database/Beam/Postgres/Full.hs
+++ b/beam-postgres/Database/Beam/Postgres/Full.hs
@@ -265,9 +265,18 @@ pgSelecting = pgSelectingWith PgCteDefault
 -- semantically valid, for example for a recursive query or a query containing
 -- volatile functions.
 --
--- Beam names every projected CTE column. A zero-column projection would make
--- that generated alias list @()@, which PostgreSQL rejects, so this function
--- rejects it before the SQL is sent.
+-- A projection with no fields is represented by omitting the CTE column-alias
+-- list. PostgreSQL then treats the CTE as a degree-zero relation: it has no
+-- columns, but it retains the row cardinality of @query@. For example, a query
+-- which produces two empty rows has the following shape:
+--
+-- @
+-- WITH "cte0" AS MATERIALIZED (SELECT FROM ...)
+-- SELECT FROM "cte0" AS "t0"
+-- @
+--
+-- Reusing such a CTE remains meaningful in joins, @EXISTS@, and aggregates
+-- even though no value can be projected from an individual row.
 pgSelectingWith
   :: forall res db placement
    . ( Projectible Postgres res
@@ -278,12 +287,10 @@ pgSelectingWith
 pgSelectingWith materialization q = do
   tblNm <- pgRegisterCte $ \name ->
     let (_ :: res, fields) = mkFieldNames @Postgres (qualifiedField name)
-    in pgOutputCteSyntax
-         "Database.Beam.Postgres.Full.pgSelectingWith"
-         name
-         fields
-         materialization
-         (fromPgSelect (buildSqlQuery (name <> "_") q))
+        body = fromPgSelect (buildSqlQuery (name <> "_") q)
+    in case nonEmpty fields of
+         Nothing -> pgCteSyntax name Nothing materialization body
+         Just fields' -> pgOutputCteSyntax name fields' materialization body
   pure (CTE.reusableForCTE tblNm)
 
 -- | An explicit lock against some tables. You can create a value of this type using the 'locked_'
@@ -446,7 +453,8 @@ insertReturning (DatabaseEntity tbl@(DatabaseTable {}))
 -- @
 --
 -- Empty insert values register no CTE. The result is still conservatively
--- 'PgCteTopLevelOnly', because placement cannot depend on a runtime value.
+-- 'PgCteTopLevelOnly', because the placement index cannot vary with the
+-- supplied values.
 cteInsert
   :: DatabaseEntity Postgres db (TableEntity table)
   -> SqlInsertValues Postgres (table (QExpr Postgres s))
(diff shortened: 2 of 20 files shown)


beam-core/Database/Beam/Backend/SQL/SQL99.hs
```diff
@@ -67,3 +68,19 @@ class IsSql99CommonTableExpressionSyntax syntax where
...
+-- supplies the CTE name and output column names around a backend-specific
+-- statement such as @DELETE ... RETURNING ...@.
+class IsSql99CommonTableExpressionSyntax syntax
+    => IsSql99DataModifyingCommonTableExpressionSyntax syntax where
+
+  -- | Backend-specific syntax for the statement inside the CTE body.
+  type Sql99CTEDataModifyingSyntax syntax :: Type
+
+  -- | Wrap a data-modifying statement as one named CTE definition.
+  cteDataModifyingSyntax :: Text -> [Text] -> Sql99CTEDataModifyingSyntax syntax -> syntax
```
LaurentRDC: As far as I can tell, data-modifying CTEs are a not part of the SQL-1999 standard at all. They are a PostgreSQL-only extension.

As such, I don't expect `beam-core` to be affected by this pull request

kushagarr: So what is it that you are suggesting ? Move entire data-modifying CTE from beam-core to beam-postgres ? and leave beam-core untouched.

LaurentRDC: Yes that's my suggestion:

* Move all of this functionality in `beam-postgres`. It should probably be exported out of `Database.Beam.Postgres.Full`
* No other package should be touched.

kushagarr: I have made the changes and updated the top level summary comment to reflect the changes as well.
I have also added support for Materialized and Not Materialised cte syntax as well.
Please review.
Thanks

LaurentRDC: Thanks, the shape of this PR looks great. I'll review the details as soon as I can

beam-postgres/test/Database/Beam/Postgres/Test/CTE.hs
```diff
@@ -0,0 +1,1102 @@
...
+  , testTopLevelOnlyDmlConsumerRendering
+  , testEmptyDmlConsumers
+  , testReturningAfterDmlConsumers
+  ]
+
+-- These tests force expressions compiled with deferred type errors in the
+-- isolated negative-fixture module. Checking fragments of GHC's error ensures
+-- an unrelated deferred error cannot make a test pass accidentally.
+typeSafetyTests :: TestTree
+typeSafetyTests = testGroup "Common table expression type-safety tests"
```
LaurentRDC, line 120: What a wonderfully clever idea!

docs/user-guide/backends/beam-postgres.md
```diff
@@ -276,24 +276,146 @@ runInsert $
...
+
+```sql
+WITH "cte0"("res0", "res1") AS (SELECT ...)
+INSERT INTO "archive_customers"("id", "name")
+SELECT "t0"."res0", "t0"."res1" FROM "cte0" AS "t0"
+```
+
+An empty CTE insert or identity CTE update registers no definition. An empty terminal insert or
+identity terminal update remains a no-op: PostgreSQL cannot execute a bare `WITH` block, so CTE
+bodies accumulated before that missing terminal statement are not executed.
```
LaurentRDC, line 477: Can you make use of the existing system to run the queries and automatically generate the SQL and results? This is done with the annotations like 

````markdown
```haskell
!example ...
...
```
````

File: beam-postgres/Database/Beam/Postgres/Syntax.hs [hunk 4]
```diff
@@ -272,9 +273,29 @@ data PgOrderingSyntax = PgOrderingSyntax { pgOrderingSyntax :: PgSyntax, pgOrder
...
 
+-- | Prefix a PostgreSQL statement with a common-table-expression list.
+-- The public CTE consumers use this prefix before their @SELECT@, @INSERT@,
+-- @UPDATE@, and @DELETE@ terminal statements, so this operation works on the
+-- shared raw syntax instead of giving the terminal statement a misleading
+-- type.
+--
+-- An empty list leaves the statement unchanged. The boolean selects
+-- @WITH RECURSIVE@ when the CTE builder used recursive bindings.
+pgWithSyntax :: Bool -> [PgCommonTableExpressionSyntax] -> PgSyntax -> PgSyntax
```
LaurentRDC: Nit: I would prefer a sum type like `data Recursion = Recursive | NonRecursive` rather than `Bool`

beam-postgres/Database/Beam/Postgres/Full.hs
```diff
@@ -316,6 +644,242 @@ pgSelectWith (CTE.With mkQ) =
...
+  :: String
+  -> Text
+  -> [Text]
+  -> PgCteMaterialization
+  -> PgSyntax
+  -> PgCommonTableExpressionSyntax
+pgOutputCteSyntax origin name fields materialization body =
+  case nonEmpty fields of
+    Nothing -> error (origin ++ ": a PostgreSQL CTE must project at least one column")
+    Just fields' -> pgCteSyntax name (Just fields') materialization body
```
LaurentRDC: Let's make use of `NonEmpty Text` rather than `[Text]` for fields. The Beam project has had other problems with expecting fields and using `error`.

This is going to impact other use-sites of `pgOutputCteSyntax`, surely

docs/user-guide/backends/beam-postgres.md
kushagarr: done

beam-postgres/Database/Beam/Postgres/Syntax.hs
kushagarr: done

beam-postgres/Database/Beam/Postgres/Full.hs
kushagarr: this is also done, but it introduced some other considerations which are also taken care of now.

beam-postgres/beam-docs.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 . ${BEAM_DOCS_LIBRARY}
 
-CHINOOK_POSTGRES_URL="https://raw.githubusercontent.com/lerocha/chinook-database/master/ChinookDatabase/DataSources/Chinook_PostgreSql.sql"
+CHINOOK_POSTGRES_URL="https://raw.githubusercontent.com/lerocha/chinook-database/1b6138b3f70a8090db48011ff4abc0e3627b22b6/ChinookDatabase/DataSources/Chinook_PostgreSql.sql"
```
LaurentRDC: Why this specific commit instead of the most recent one, `7f67772503d71ba90f19283c38e93923addb43fa`?

beam-postgres/beam-postgres.cabal
LaurentRDC, line 1: Please increase the version number. Since this isn't a breaking change, you can increment to 0.6.3.0

beam-postgres/Database/Beam/Postgres/Full.hs
LaurentRDC, line 1: Please make sure to add a `@since 0.6.3.0` annotation in the docstrings to all new functions

LaurentRDC:
We're getting really close!

beam-postgres/beam-docs.sh
kushagarr: I pinned `1b6138b3f70a...` because its PostgreSQL fixture matches the existing checksum and the legacy CamelCase schema used by `Chinook.Schema`.

The fixture at `7f6777...` is not a drop-in replacement: it creates and connects to its own database and uses snake_case identifiers, whereas `beam-docs.sh` loads the fixture into a temporary database and expects names such as `"Invoice"` and `"InvoiceId"`.

I see three options:

1. Keep the compatible pin and add a comment explaining it.
2. Revert this unrelated change from the CTE PR.
3. Migrate the documentation setup and schema to the current Chinook fixture in a separate change.

My preference is to avoid bundling that migration into this PR, but I’m happy to follow whichever approach you prefer.

beam-postgres/Database/Beam/Postgres/Full.hs
kushagarr: done

beam-postgres/beam-postgres.cabal
kushagarr: done

beam-postgres/beam-docs.sh
kushagarr: One additional detail: the current Nix documentation CI pre-populates the Chinook cache from `docs/data/schemas/Chinook_PostgreSql.sql`, so `beam-docs.sh` does not use either the download URL or its checksum during CI.

The pin only affects clean, non-Nix documentation builds where that cached fixture is absent. Given that, I’m also happy to revert this one-line change and keep the CTE PR focused.

kushagarr: I have reverted it..one should look at it separately, not in this PR.

kushagarr:
It closes https://github.com/haskell-beam/beam/issues/290

LaurentRDC (approved):
Awesome, thanks @kushagarr!

LaurentRDC:
@kushagarr released in [beam-postgres 0.6.3.0](https://hackage.haskell.org/package/beam-postgres-0.6.3.0)

PR #826, opened 2026-07-26
docs: document beam-large-records

### LINKED ISSUE #825: Document the existence and purpose of `beam-large-records`
[`beam-large-records`](https://github.com/well-typed/large-records) exists to speed up compilation of beam schemas with lots of fields. We should mention its existence and purpose in the documentation!

ColumbusLabs:
## Summary

- document when `beam-large-records` is useful for schemas with many fields
- show the plugin, integration import, and `largeRecord` annotation
- list the direct dependencies required by the plugin and link authoritative package documentation

Closes #825.

## Validation

- verified the package metadata and example against the official `beam-large-records` source
- verified the quoted benchmark values against the upstream timing data and methodology
- verified all three added links resolve
- checked Markdown fence balance
- `git diff --check`

The repository's `nix build .#docs -L` check was not available locally because Nix is not installed.

### DIFF
--- a/docs/user-guide/models.md
+++ b/docs/user-guide/models.md
@@ -9,6 +9,29 @@ type with a `Beamable` instance) or an explicit column. A column is specified
 using the `Columnar` type family applied to the type's parameter and the
 underlying Haskell type of the field.
 
+## Schemas with many fields
+
+For schemas with many fields, the separate
+[`beam-large-records`](https://github.com/well-typed/large-records/tree/main/beam-large-records)
+package can reduce compilation time. It integrates Beam with
+`large-records` so that records marked with
+[`largeRecord`](https://hackage.haskell.org/package/large-records/docs/Data-Record-Plugin.html#v:largeRecord)
+can derive `Beamable`.
+
+Add `large-records`, `ghc-prim`, `large-generics`, `record-hasfield`, and
+`beam-large-records` to your build dependencies. Then enable the `large-records`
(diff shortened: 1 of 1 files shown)


docs/user-guide/models.md
```diff
@@ -9,6 +9,29 @@ type with a `Beamable` instance) or an explicit column. A column is specified
 using the `Columnar` type family applied to the type's parameter and the
 underlying Haskell type of the field.
 
+## Schemas with many fields
+
+For schemas with many fields, the separate
+[`beam-large-records`](https://github.com/well-typed/large-records/tree/main/beam-large-records)
+package can reduce compilation time. It integrates Beam with
```
LaurentRDC: Is it possible to be more quantitative here? For example, how many fields is too many, and how much can it reduce compilation time?

ColumbusLabs: Good question. I couldn’t find a defensible universal field-count cutoff, but the upstream synthetic benchmark provides useful context: with GHC 8.8.4 and `-O0`, its 20-field module compiled in 318 ms before and 80 ms with `large-records`—about 75% less. At 100 fields, the corresponding times were 2.28 s and 444 ms—about 81% less. Since that benchmark is not Beam-specific, I’ve updated the docs to present those as illustrative measurements and recommend profiling the actual schema rather than treating 20 fields as a threshold.

LaurentRDC (approved):
Thank you!

PR #827, opened 2026-08-05
Add PostgreSQL partial-column inserts with composable CTE support

kushagarr:
## Summary

Adds PostgreSQL-specific partial-column inserts that preserve `ON CONFLICT` support and compose cleanly with `RETURNING` and data-modifying CTEs.

## Motivation

The existing PostgreSQL insert helpers support `insertValues`, `insertExpressions`, and `insertFrom`, but require values for every table column. This prevents CTE inserts from omitting generated, identity, or database-defaulted columns.

Beam’s backend-independent `insertOnly` supports partial-column inserts, but cannot express PostgreSQL’s `ON CONFLICT` behavior. Beam also provides backend-independent conflict handling through [`BeamHasInsertOnConflict`](https://hackage.haskell.org/package/beam-core/docs/Database-Beam-Backend-SQL-BeamExtensions.html#t:BeamHasInsertOnConflict), whose `insertOnConflict` accepts full-table values. `pgInsertOnly` adds partial-target selection with conflict handling.

For query-backed inserts, `default_` cannot substitute for omitted columns: PostgreSQL rejects `SELECT DEFAULT, source.name FROM source`, while `INSERT INTO users (name) SELECT source.name FROM source` lets the omitted column use its default.

## Changes

- Add `pgInsertOnly` for partial-column PostgreSQL inserts with `PgInsertOnConflict`.
- Add `cteInsertCommand` to lift any `SqlInsert Postgres table` into a side-effect-only CTE.
- Add `cteInsertCommandReturning` to attach a `RETURNING` projection and expose the inserted rows as a reusable CTE.
- Refactor the existing full-row insert helpers through the same implementation.
- Preserve the existing `insert`, `insertReturning`, `cteInsert`, and `cteInsertReturning` APIs.
- Document the new composition model and record the APIs for `beam-postgres-0.6.4.0`.
- Allow the integration suite to use an externally managed PostgreSQL instance through `BEAM_POSTGRES_TEST_CONNSTR`.

## Design

Partial-column selection is treated as a property of insert construction, while conflict handling, `RETURNING`, and CTE placement remain independently composable:

~~~haskell
command =
  Pg.pgInsertOnly table selectedFields values onConflict

runInsert command
returning command resultProjection
Pg.cteInsertCommand command
Pg.cteInsertCommandReturning command resultProjection
~~~

`pgInsertOnly` delegates target-column projection and base SQL generation to Beam’s existing `insertOnly` implementation, then adds PostgreSQL’s conflict clause. This avoids maintaining a second insert renderer and keeps column ordering, quoting, empty-input behavior, and type checking aligned with the backend-independent implementation.

The command-level CTE adapters also avoid introducing narrowly specialized partial-insert CTE APIs.

## Compatibility

This is additive and does not require changes to existing user code. Existing PostgreSQL insert and CTE helpers retain their signatures and behavior.

`cteInsertCommandReturning` returns `Nothing` only for `SqlInsertNoRows`. A valid `ON CONFLICT DO NOTHING` command still returns a reusable relation, which may contain no rows after execution.

For both command adapters, `SqlInsert` does not carry the target table’s database type, so that type is not tied to the enclosing `PgWith` block.

## Validation

- Full `beam-postgres` test suite: 149 tests passed against PostgreSQL 18.4 with GHC 9.6.7.
- Added rendering and integration coverage for:
  - selected-column ordering;
  - `insertData` and `insertFrom`;
  - generated identities and database defaults;
  - `ON CONFLICT DO NOTHING`;
  - `ON CONFLICT DO UPDATE` and `excluded` values;
  - SQL parameter ordering;
  - side-effect-only and returning CTEs;
  - empty inserts and zero-column projections;
  - compatibility between existing helpers and the composable command path.
- Added negative type tests for invalid CTE placement and mismatched insert projections.
- Haddock generation completed successfully.
- `git diff --check` passes.

### DIFF
--- a/beam-postgres/Database/Beam/Postgres/Full.hs
+++ b/beam-postgres/Database/Beam/Postgres/Full.hs
@@ -265,9 +265,18 @@ pgSelecting = pgSelectingWith PgCteDefault
 -- semantically valid, for example for a recursive query or a query containing
 -- volatile functions.
 --
--- Beam names every projected CTE column. A zero-column projection would make
--- that generated alias list @()@, which PostgreSQL rejects, so this function
--- rejects it before the SQL is sent.
+-- A projection with no fields is represented by omitting the CTE column-alias
+-- list. PostgreSQL then treats the CTE as a degree-zero relation: it has no
+-- columns, but it retains the row cardinality of @query@. For example, a query
+-- which produces two empty rows has the following shape:
+--
+-- @
+-- WITH "cte0" AS MATERIALIZED (SELECT FROM ...)
+-- SELECT FROM "cte0" AS "t0"
+-- @
+--
+-- Reusing such a CTE remains meaningful in joins, @EXISTS@, and aggregates
+-- even though no value can be projected from an individual row.
 pgSelectingWith
   :: forall res db placement
    . ( Projectible Postgres res
@@ -278,12 +287,10 @@ pgSelectingWith
 pgSelectingWith materialization q = do
   tblNm <- pgRegisterCte $ \name ->
     let (_ :: res, fields) = mkFieldNames @Postgres (qualifiedField name)
-    in pgOutputCteSyntax
-         "Database.Beam.Postgres.Full.pgSelectingWith"
-         name
-         fields
-         materialization
-         (fromPgSelect (buildSqlQuery (name <> "_") q))
+        body = fromPgSelect (buildSqlQuery (name <> "_") q)
+    in case nonEmpty fields of
+         Nothing -> pgCteSyntax name Nothing materialization body
+         Just fields' -> pgOutputCteSyntax name fields' materialization body
   pure (CTE.reusableForCTE tblNm)
 
 -- | An explicit lock against some tables. You can create a value of this type using the 'locked_'
@@ -446,7 +453,8 @@ insertReturning (DatabaseEntity tbl@(DatabaseTable {}))
 -- @
 --
 -- Empty insert values register no CTE. The result is still conservatively
--- 'PgCteTopLevelOnly', because placement cannot depend on a runtime value.
+-- 'PgCteTopLevelOnly', because the placement index cannot vary with the
+-- supplied values.
 cteInsert
   :: DatabaseEntity Postgres db (TableEntity table)
   -> SqlInsertValues Postgres (table (QExpr Postgres s))
@@ -482,9 +490,25 @@ cteInsert table values onConflict_ =
 -- The generated statement has the shape:
 --
 -- @
--- WITH cte0 AS (INSERT INTO users ... RETURNING ...)
--- SELECT ... FROM cte0
+-- WITH "cte0"("res0", ...) AS
+--        (INSERT INTO "users" ... RETURNING ...)
+-- SELECT ... FROM "cte0" AS "t0"
 -- @
+--
+-- The projection may contain no fields. In that case Beam preserves one
+-- degree-zero result row per inserted row. PostgreSQL requires at least one
+-- @RETURNING@ expression, so the CTE contains a private boolean sentinel while
+-- the final SELECT exposes no columns:
+--
+-- @
+-- WITH "cte0"("res0") AS
+--        (INSERT INTO "users" ... RETURNING NULL::boolean)
+-- SELECT FROM "cte0" AS "t0"
+-- @
+--
+-- The sentinel is not part of the returned Haskell value. If neither the final
+-- statement nor another CTE needs the inserted-row output, prefer 'cteInsert'.
+-- It omits @RETURNING@ and produces no reusable result.
 cteInsertReturning
   :: ( Projectible Postgres a
      , ThreadRewritable PostgresInaccessible a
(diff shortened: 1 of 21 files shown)


beam-postgres/Database/Beam/Postgres/Full.hs
```diff
@@ -402,13 +404,35 @@ insert :: DatabaseEntity Postgres db (TableEntity table)
...
+-- rendering; this function adds PostgreSQL's @ON CONFLICT@ extension.
+--
+-- @since 0.6.3.1
+pgInsertOnly
+  :: ProjectibleWithPredicate AnyType () Text (QExprToField r)
+  => DatabaseEntity Postgres db (TableEntity table)
+  -> (table (QField s) -> QExprToField r)
+  -> SqlInsertValues Postgres r
+  -> PgInsertOnConflict table
+  -> SqlInsert Postgres table
```
LaurentRDC, line 422: How is this different than using `insert` in combination with `default_` for the fields you don't want to specify?

kushagarr: Query Based inserts is the primary reason, for example

a valid beam code can generate both of the below queries

```
-- Valid
INSERT INTO users (id, name)
VALUES (DEFAULT, 'Alice');

-- Invalid
INSERT INTO users (id, name)
SELECT DEFAULT, source.name FROM source;
```

the invalid one fails at runtime, Default not allowed inside select.

A partial target is able to solve this

```
INSERT INTO users (name)
SELECT source.name FROM source;
```

The above is the main reason,
there are other secondary ergonomics which could be catered to here, not new insert capability

- insertValues contains ordinary Haskell records and cannot represent default_.
- A caller must switch to insertExpressions and build an entire table-shaped expression merely to put default_ in omitted positions.
- insertData naturally supports the smaller projected shape with insertOnly.
- The partial shape makes it explicit which columns the application controls.
- It matches Beam-core’s existing insertOnly abstraction.
- pgInsertOnly combines that abstraction with PostgreSQL ON CONFLICT, which Beam-core’s insertOnly cannot provide.
- also there are queries with column level permissions where say a generated column is not granted an insert permission so a default_ also will not be accepted for such a column in a table.

kushagarr: @LaurentRDC any thoughts ?

LaurentRDC: I was on vacation. Taking another look now!

beam-postgres/Database/Beam/Postgres/Full.hs
```diff
@@ -402,13 +404,35 @@ insert :: DatabaseEntity Postgres db (TableEntity table)
...
-      SqlInsertNoRows
+insert tbl@(DatabaseEntity (DatabaseTable {})) values =
+  pgInsertOnly tbl id values
+
+-- | Build a PostgreSQL insert over a caller-selected subset of table fields,
+-- with support for 'PgInsertOnConflict'. Beam's backend-independent
+-- 'Beam.insertOnly' owns the typed field projection and base @INSERT@
+-- rendering; this function adds PostgreSQL's @ON CONFLICT@ extension.
+--
+-- @since 0.6.3.1
```
LaurentRDC: Might as well mention Beam's backend-independent support of `ON CONFLICT`: https://hackage-content.haskell.org/package/beam-core-0.11.1.0/docs/Database-Beam-Backend-SQL-BeamExtensions.html#t:BeamHasInsertOnConflict

LaurentRDC:
Looks good to me overall. 
I've marked this ready-for-review because I thought that was needed to run CI, but that's not true. So let me know if you plan on making further changes

beam-postgres/Database/Beam/Postgres/Full.hs
kushagarr: Added the `BeamHasInsertOnConflict` link to the Haddock and user guide, explaining that `insertOnConflict` accepts full-table values while `pgInsertOnly` adds partial-target selection with conflict handling. The guide also includes the `INSERT … SELECT` example showing why `default_` cannot replace omitted columns.

The two new adapters are now `cteInsertCommand` and `cteInsertCommandReturning` (formerly `pgCteInsert` and `pgCteInsertReturning`), and both document the database-type limitation of `SqlInsert`. Released signatures are unchanged; release metadata is aligned to `0.6.4.0`.

Validated commit 391d152 with GHC 9.6.7: library build and Haddocks passed, generated links checked, and all 149 tests passed against PostgreSQL 18.4, including rendering, negative-type, and legacy-wrapper compatibility tests. Stale-name/version and whitespace checks pass. The unchanged [GHC CI matrix](https://github.com/haskell-beam/beam/actions/runs/35700775592) is running.

kushagarr:
Thanks for the review — pushed a finishing pass, and I'm not planning further changes.

**Conflict-support docs** (your comment on `pgInsertOnly`): added the reference to
`BeamHasInsertOnConflict` in both the Haddock and the user guide, with the distinction
made explicit — `insertOnConflict` accepts full-table `SqlInsertValues`, so it can't
express a partial target.

**Renamed the new CTE adapters**: `pgCteInsert` → `cteInsertCommand` and
`pgCteInsertReturning` → `cteInsertCommandReturning`. What separates them from
`cteInsert`/`cteInsertReturning` isn't that they're Postgres-specific —  it's that they take an already-built `SqlInsert` instead of table + values + conflict action. The `pg` prefix didn't convey that; `Command` does. It also keeps
`pgCteInsert` available in case the existing names are ever normalised.

**Documented a limitation**: `SqlInsert` doesn't carry the target table's database type, so
for the command-level adapters that type isn't tied to the enclosing `PgWith` block. Noted
on both functions and in the user guide. The existing `cteInsert`/`cteInsertReturning`
wrappers keep their signatures and their `db` linkage.

**Version**: corrected `0.6.3.1` → `0.6.4.0`. This release only adds bindings. Updated the
cabal version, the ChangeLog and the `@since` annotations to match.

Existing exports and behaviour are unchanged.

If it looks good to you we can merge it and make a new release please

